PR #5, opened 2020-09-17
Fix CI

marcellp:


ivababukova (approved):
wow, how did that ever work before that

PR #6, opened 2020-09-18
Fix worker to use InfraMock endpoints for S3.

marcellp:


src/helpers/load_count_matrix.py
```diff
@@ -41,5 +41,6 @@ def _load_file(matrix_path):
 def get_adata(adata, experiment_id):
     print(datetime.datetime.utcnow(), "adata does not exist, I need to download it ...")
     matrix_path = get_item_from_dynamo(experiment_id, "matrixPath")
+    print(matrix_path, "is our path!!")
```
ivababukova: remove this print

ivababukova (approved):

PR #9, opened 2020-10-01
Add storage class and volume cleanup job to the worker

marcellp:


chart/templates/worker-spawn-right-to-api.yaml
```diff
@@ -7,6 +7,9 @@ rules:
 - apiGroups: ["batch"]
   resources: ["jobs"]
   verbs: ["create"]
+- apiGroups: [""]
+  resources: ["persistentvolumeclaims"]
```
ivababukova, line 11: we will still need to handle the actual creation of this from the API code, right?

chart/templates/unused-volume-cleanup.yaml
```diff
@@ -0,0 +1,66 @@
...
+          serviceAccountName: unused-volume-cleanup
+          restartPolicy: OnFailure
+          containers:
+          - name: unused-volume-cleanup
+            image: dtzar/helm-kubectl:3.3.1
+            args:
+            - /bin/sh
+            - -c
+            - |
+              kubectl describe -n {{ .Release.Namespace }} pvc | grep -E "^Name:.*$|^Namespace:.*$|^Mounted By:.*$" | grep -B 2 "<none>" | grep -E "^Name:.*$|^Namespace:.*$" | cut -f2 -d: | paste -d " " - - | xargs -n2 bash -c 'kubectl -n ${1} delete pvc ${0}' || true
```
ivababukova, line 22: would that script hang if there are pvc that are currently in use and make them stuck in `terminating` state?

ivababukova (approved):

PR #11, opened 2020-10-08
Implement R container

marcellp:


python/README.md
```diff
@@ -0,0 +1,129 @@
...
+# Development
+
+## Run locally
+
+### 1. Set up environment
+To run this code locally, first make sure you start a python virtual environment, activate it and install all the requirements:
+
+        python3.7 -m venv
+        source venv/bin/activate
+        pip install -r requirements.txt
```
ivababukova: delete that

ivababukova: put instructions for the vscode way to run it instead

python/README.md
```diff
@@ -0,0 +1,129 @@
...
+            "sourceMatrixPath": "pbmc_count_matrices/hg19/"
+        }
+    }
+
+
+### 3. Run the code
+To run the worker locally, execute the following commands in a terminal:
+
+    cd src/
+    python work.py
```
ivababukova: this needs to be updated too

python/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+Note that the worker will automatically switch itself off if it doesn't receive any tasks for 20 minutes. In this case, simply rerun it again and it will start, as normal.
+
+## Run tests
+Go to the src/ and run:
+
+    python3 -m pytest --cov=.
```
ivababukova: say that the environment has to be set to test

ivababukova (approved):

PR #27, opened 2020-10-19
Download files to /data in disk

ivababukova:
# Changes:
1. When the worker is started, downloads all files in S3 under the experiment id folder. Reads correct experiment id from an environment variable.
2. Removed unused PrepareExperiment task
3. Before each task is done, checks if the file has been changed by comparing the S3 eTag with the file md5 checksum. If the file is changed, downloads all files again.
4. Not reading `count_matrix` entry in DynamoDB anymore.

# Still left to do:
1. Write unit tests.
2. Merge only after all anndata and R files have been uploaded under folders in s3 with the corresponding experiment ID.
3. Raise a merge request with the inframock changes required to keep this running fine locally.


python/src/helpers/count_matrix.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    s3_etag = resp["ETag"]
+    file_etag = _calculate_file_etag(file_path)
+
+    return s3_etag != file_etag
+
+
+def _create_base_path_dirs():
+    if not os.path.exists(PATH_TO_FILES):
+        print("not existing, creating: ")
```
ivababukova: remove this print

python/src/tasks/tasks.py
```diff
@@ -41,31 +53,36 @@ def submit(self, msg, adata):
                         error=True,
                     )
                 ]
-            return result, adata
+            return result
+
+    def _factory(self, msg):
+        print("before checking for change: ", self.adata_path)
```
marcellp: remove this too

python/src/tasks/tasks.py
```diff
@@ -41,31 +53,36 @@ def submit(self, msg, adata):
...
+        if count_matrix.is_file_changed(self.adata_path):
+            print("There has been recent write to the file, have to download it again.")
+            # TODO: after introducing multiple-sample support, make this
+            #  more efficient by making it able to download only a specific file too.
+            count_matrix.download_files()
+            self._initialise_adata()
```
marcellp: I think this code overall is still too difficult to read, but I think now reading all the code I have a better understanding of what you were doing. Here is a possibly better way to go, and this lends itself nicely to the R development that is going to be carried out immediately, as well as multiple experiment support. Neither is supported very well as the code currently stands.

Expose just one function, `download_files()` from `count_matrix`. It does the following:

1. check if the path `/data/$EXPERIMENT_ID` exists, save it to a variable called `path_exists`.
2. call boto with `list_objects`, save it to a variable called `objects`.
3. for each object in `objects`, call `_calculate_file_etag()`. write it so it handles files not existing, and in that case return `None` if return value of `_calculate_file_etag()` does not match, call `download_file(object_name)`, which will download the file from the object path `213213asdas2134213/python.h5ad` into `/data/213213asdas2134213/python.h5ad`, as appropriate.
4. the etags computed should be saved into a json and referenced instead of recomputing if possible

we can jump on a call and I can show you what i mean to discuss this.

marcellp (approved):
lgtm

PR #28, opened 2020-10-22
Have R perform differential expression

marcellp:


.github/workflows/ci.yaml
```diff
@@ -52,6 +52,7 @@ jobs:
       name: Run unit tests
       run: |-
         cd src/
+        gunzip -k ../../data/test/python.h5ad.gz
         CLUSTER_ENV="test" python -m pytest --cov=.
```
ivababukova, line 56: why `guzip`-ping?

ivababukova: it's fine, I found the answer

python/src/tests/de_result.json
```diff
@@ -0,0 +1 @@
...", "ENSG00000171863", "ENSG00000181163", "ENSG00000109971", "ENSG00000105373", "ENSG00000145425", "ENSG00000108107", "ENSG00000174444", "ENSG00000233927", "ENSG00000104964", "ENSG00000142676", "ENSG00000161016", "ENSG00000124614", "ENSG00000172809", "ENSG00000170889", "ENSG00000114942", "ENSG00000227507", "ENSG00000144713", "ENSG00000251562", "ENSG00000197958", "ENSG00000134419", "ENSG00000174748", "ENSG00000112306", "ENSG00000122026", "ENSG00000188846", "ENSG00000089157", "ENSG00000162244", "ENSG00000122406", "ENSG00000142937", "ENSG00000198034", "ENSG00000164587", "ENSG00000177600", "ENSG00000196405", "ENSG00000156508", "ENSG00000186468", "ENSG00000063177", "ENSG00000105640", "ENSG00000130255", "ENSG00000105193", "ENSG00000143947", "ENSG00000145592", "ENSG00000115268", "ENSG00000118181", "ENSG00000142541", "ENSG00000171858", "ENSG00000136942", "ENSG00000148303", "ENSG00000167526", "ENSG00000108298", "ENSG00000198755", "ENSG00000229117", "ENSG00000198242", "ENSG00000137154", "ENSG00000083845", "ENSG00000140988", "ENSG00000147403", "ENSG00000197756", "ENSG00000231500", "ENSG00000213741", "ENSG00000149273", "ENSG00000100316", "ENSG00000177954", "ENSG00000168028", "ENSG00000105372"]}
```
ivababukova: do we need this file in github?

python/src/tasks/differential_expression.py
```diff
@@ -62,26 +63,37 @@ def compute(self):
...
+        with open("/data/result.json", "w") as f:
+            json.dump(r.json(), f)
+
+        result = pandas.DataFrame.from_dict(r.json())
+        result = result[["Z", "M", "Gene"]]
+        result = result.rename(
+            columns={"Z": "zscore", "M": "log2fc", "Gene": "gene_names"}
+        )
+
+        result["abszscore"] = result["zscore"].abs()
```
ivababukova: could we not get `"Z"`, `"W"`, and `"Gene"` from R and then Python just sends them back to the API directly so that we don't do those transformations here?

ivababukova: If we were to do the transformations from Python, then maybe it is worth deleting the `result.json` file or using `tempFile` to create it, since we don't need it after the `result` object is constructed.

.github/workflows/ci.yaml
```diff
@@ -52,6 +52,7 @@ jobs:
       name: Run unit tests
       run: |-
         cd src/
+        gunzip -k ../../data/test/python.h5ad.gz
```
ivababukova, line 55: do we not need to change inframock as well to download the files using the updated file path?

r/src/work.r
```diff
@@ -1,13 +1,103 @@
...
-        response$set_body("up")
+    loaded <- F
+    data <- NULL
+
+    while (!loaded) {
+        data <- tryCatch(
+            {
+                f <- readRDS(
+                    paste(
+                        "/data", "5e959f9c9f4b120771249001", "rb.rds",
```
ivababukova: hardcoded experiment id in here

python/src/tests/de_result.json
marcellp: Yes, it is for mocking the R worker's DE result.

python/src/tasks/differential_expression.py
marcellp: Sure, fixed.

.github/workflows/ci.yaml
marcellp: Yes, as I said in the standup. Will update InfraMock after this merge.

r/src/work.r
marcellp: True, fixed.

ivababukova (approved):
that's much better, good to go!

PR #31, opened 2020-10-29
Increase worker timeout to 1 hour

ivababukova:


marcellp:
I see what you were thinking about not having a check when sending stuff back on the API.

The reason I thought this was because the message in `consume_message.py` was wrong. It should be

> Skipping processing task with uuid

instead of

> Skipping sending task with uuid

marcellp (changes requested):
I see what you were thinking about not having a check when sending stuff back on the API.

The reason I thought this was because the message in `consume_message.py` was wrong. It should be

> Skipping processing task with uuid

instead of

> Skipping sending task with uuid

marcellp (approved):

PR #33, opened 2020-11-02
Create a deployment file for master staging deployments

marcellp:


.github/workflows/ci.yaml
```diff
@@ -274,7 +289,7 @@ jobs:
 
       - name: Change name of deployment file for staging deployment
         if:
-          github.event_name == 'pull_request' && matrix.environment == 'staging'
+          matrix.environment == 'staging'
```
marcellp, line 292: The idea is to also push to `staging_candidates/` a release file when the push to master happens. This is so we can more easily create the deployment from the script when needed.

.github/workflows/ci.yaml
```diff
@@ -209,31 +209,46 @@ jobs:
...
-          | yq w -d0 - metadata.name "$DEPLOYMENT_NAME-$REF_DASHED" \
+          | yq w -d1 - spec.values.images.r "$IMAGE_NAME-r" \
+          | yq w -d1 - spec.values.images.python "$IMAGE_NAME-python" \
           > $DEPLOYMENT_NAME.yaml
 
           cat $DEPLOYMENT_NAME.yaml
+        env:
+          IMAGE_NAME: ${{ format('{0}/{1}:{2}', steps.login-ecr.outputs.registry, needs.build-docker.outputs.repo-name, needs.build-docker.outputs.tag) }}
 
+      # TODO: to be deprecated
```
ivababukova, line 237: why is this going to be deprecated?

.github/workflows/ci.yaml
```diff
@@ -209,31 +209,46 @@ jobs:
...
 
+          if [ "${{ matrix.environment }}" = "production" ]; then
+            SANDBOX_ID="default"
+          fi
+
+          if [ "${{ matrix.environment }}" = "staging" ]; then
+            SANDBOX_ID="STAGING_SANDBOX_ID"
+          fi
+
+          echo "::set-output name=sandbox-id::$SANDBOX_ID"
```
marcellp, line 220: The idea is to change the names of the namespaces/deployments from `worker-refs-heads-master` to `worker-sandboxId`, which is `default` by default and a hash for a deployment created with the script in the future.

chart-infra/templates/image-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
```
ivababukova, line 2: what is this ConfigMap for?

.github/workflows/ci.yaml
marcellp: The chart will have the image names in it as a ConfigFile.

Right now there can only be one `worker_instance.yaml` file for the entire repo, this is not great, because we can't have a separate one for each staging environment or even staging vs prod.  This is the file that the API reads and fills in to create a new worker instance.

Now instead of reading from that file, the API will be able to read the image IDs directly from [here](https://github.com/biomage-ltd/iac/blob/master/releases/production/worker.yaml) (under `specs.images.python` and `specs.images.r`) instead of a separate file. This also versions them per deployment which is great. The idea is that because the API will be aware of its sandbox ID through environment variables (which I think it already is), it will be able to automatically know the worker namespace to be `worker-sandboxId`.

For now, this is transparent to the API as both the `worker_instance.yaml` as well as the API infrastructure deployment will change the name from `worker-refs-heads-master` to `worker-default`, which means there's no downtime.

chart-infra/templates/image-configmap.yaml
marcellp: See my response above.

.github/workflows/ci.yaml
marcellp: Actually, you can even see it [here](https://github.com/biomage-ltd/iac/blob/master/releases/staging-candidates/worker/refs-pull-33-merge.yaml) now that the pipeline is done.

ivababukova: I see, and it will know which folder in `releases` it will need to go so it can read the image id directly

ivababukova (approved):

PR #44, opened 2020-11-19
Fix bug where Differential Expression tool cannot set `all` as first argument.

marcellp:


ivababukova (approved):
approving this, assuming that you have tested it end to end with inframock

marcellp:
> approving this, assuming that you have tested it end to end with inframock

still testing it and tests don't pass yet

PR #51, opened 2021-01-12
Lock pagoda version and update readme for linux users

cosa65:
# Background
When the differential expression is run, the R worker uses the pagoda function getDifferentialGenes, which seems to not be available in newer versions of pagoda, so this library should be locked to a version that does have it in order to have the differential expression run correctly.

#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-449?atlOrigin=eyJpIjoiYjNjNDQ2OGQ1YTlmNDA5ZWIyYTdiNGQzNjAyY2ZjYzEiLCJwIjoiaiJ9

Added an item to the readme with a possible fix for an issue linux users may run into.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/r/Dockerfile
+++ b/r/Dockerfile
@@ -23,7 +23,7 @@ RUN Rscript setup/install_or_die.r BiocManager
 RUN R -e 'BiocManager::install(c("AnnotationDbi", "BiocGenerics", "GO.db", "pcaMethods", "org.Dr.eg.db", "org.Hs.eg.db", "org.Mm.eg.db"))'
 
 RUN R -e 'install.packages("RcppSpdlog")'
-RUN R -e 'devtools::install_github("kharchenkolab/pagoda2")'
+RUN R -e 'devtools::install_github("kharchenkolab/pagoda2@v0.1.4")'
 RUN R -e 'devtools::install_github("kharchenkolab/conos@v1.3.1")'
 
 RUN R -e 'install.packages("gprofiler2")'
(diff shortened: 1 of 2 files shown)


r/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN Rscript setup/install_or_die.r BiocManager
 RUN R -e 'BiocManager::install(c("AnnotationDbi", "BiocGenerics", "GO.db", "pcaMethods", "org.Dr.eg.db", "org.Hs.eg.db", "org.Mm.eg.db"))'
 
 RUN R -e 'install.packages("RcppSpdlog")'
-RUN R -e 'devtools::install_github("kharchenkolab/pagoda2")'
+RUN R -e 'devtools::install_github("kharchenkolab/pagoda2@v0.1.4")'
 RUN R -e 'devtools::install_github("kharchenkolab/conos@v1.3.1")'
```
ivababukova, line 27: is it worth pre-emptively locking the other packages too?

cosa65: @marcellp might have a better idea.

marcellp: My understanding is that it's not possible to do it in the way we are installing stuff just now -- you can pin `install_github` stuff to a particular GitHub branch, but to install packages using `install.packages` you need to individually find the latest versions of each of the packages and use those through a direct URL, which I think is probably not the best thing to do straight away.

My hunch would be to have a separate ticket for investigating and implementing the best way to lock the non-GitHub packages.

ivababukova: Do you mind writing that ticket up @marcellp? Seems like you have the most context of the work needed and why we need to do it on this one.

ivababukova (approved):

PR #53, opened 2021-01-14
Multiple samples support in diffexpr

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-434?atlOrigin=eyJpIjoiN2RiNzViMzEyY2RmNDM5YzgxMTRmYTg1OTg5Y2ExZGIiLCJwIjoiaiJ9

https://ui-fd7d4etwc13uaatqge73adaewo.scp-staging.biomage.net/

# Changes
### Code changes
- Added support for basis sample filtering of differential expression sets
- Updated functionality when receiving the "rest" keyword to load all cell sets in the same hierarchy as the first one instead of all cells.
- Updated marking of cells so that those that are in the intersection of both sets are excluded from the computation since repeated cells could cause problems with the implementation of the r worker.
- Refactored tests for differential expression in python.
- Added tests for the new functionalities in differential expression.

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/python/src/helpers/find_cell_ids_in_same_hierarchy.py
+++ b/python/src/helpers/find_cell_ids_in_same_hierarchy.py
@@ -4,13 +4,14 @@ def get_all_cell_ids_in(cell_set):
     children = cell_set.get("children", None)
     if children:
         for child in children:
-            cell_ids += all_cell_ids_in(child)
+            cell_ids += get_all_cell_ids_in(child)
 
     return cell_ids
 
+
 def get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=None):
     cell_sets_to_return = []
-  
+
     for cell_set in cell_sets:
         if ignoring_set_id and cell_set["key"] == ignoring_set_id:
             continue
@@ -20,12 +21,13 @@ def get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=None):
 
     return cell_sets_to_return
 
-# returns a list with 
+
+# returns a list with
 def find_cell_ids_in_same_hierarchy(key, cell_sets):
     for cell_set in cell_sets:
         if cell_set["key"] == key:
             return get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=key)
-        
+
         if cell_set.get("children", None):
             result = find_cell_ids_in_same_hierarchy(key, cell_set["children"])
 
--- a/python/src/helpers/find_cell_ids_in_same_hierarchy.py
+++ b/python/src/helpers/find_cell_ids_in_same_hierarchy.py
@@ -1,0 +1,35 @@
+def get_all_cell_ids_in(cell_set):
+    cell_ids = cell_set["cellIds"]
+
+    children = cell_set.get("children", None)
+    if children:
+        for child in children:
+            cell_ids += all_cell_ids_in(child)
+
+    return cell_ids
+
+def get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=None):
+    cell_sets_to_return = []
+  
+    for cell_set in cell_sets:
+        if ignoring_set_id and cell_set["key"] == ignoring_set_id:
+            continue
+
+        # cell_sets_to_return.append(cell_set["key"])
+        cell_sets_to_return += get_all_cell_ids_in(cell_set)
+
+    return cell_sets_to_return
+
+# returns a list with 
+def find_cell_ids_in_same_hierarchy(key, cell_sets):
+    for cell_set in cell_sets:
+        if cell_set["key"] == key:
+            return get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=key)
+        
+        if cell_set.get("children", None):
+            result = find_cell_ids_in_same_hierarchy(key, cell_set["children"])
+
+            if result:
+                return result
+
+    return []
--- a/python/src/helpers/find_cell_ids_in_same_hierarchy.py
+++ b/python/src/helpers/find_cell_ids_in_same_hierarchy.py
@@ -16,13 +16,12 @@ def get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=None):
         if ignoring_set_id and cell_set["key"] == ignoring_set_id:
             continue
 
-        # cell_sets_to_return.append(cell_set["key"])
         cell_sets_to_return += get_all_cell_ids_in(cell_set)
 
     return cell_sets_to_return
 
 
-# returns a list with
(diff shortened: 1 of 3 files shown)


cosa65:
I am getting a warning when running the differential expression:
`A value is trying to be set on a copy of a slice from a DataFrame`, is this an issue or should I ignore it?

ivababukova:
@cosa65 if you get the same warning in the master version of the worker, I would say that it is ok.

README.md
```diff
@@ -81,3 +81,18 @@ For more details on the individual runners, check out the README files in their
...
+
+Go to `docker-compose.yml`
+In the python and r entries add at the end:
+
+```
+extra_hosts:
+      - "host.docker.internal:host-gateway"
+```
+
+IMPORTANT: Don't include this in a PR, because it will break stuff on macOS.
```
ivababukova: I am sure I saw this change added as part of a previous PR. Did you not get it merged at the end?

cosa65: Yes, but I had to start this branch from the other PR when it wasn't merge yet because it was necessary for me to be able to use the worker.

I'll rebase and fix it.

cosa65: Now it should be ok.

python/src/helpers/find_cell_ids_in_same_hierarchy.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=None):
+    cell_sets_to_return = []
+
+    for cell_set in cell_sets:
+        if ignoring_set_id and cell_set["key"] == ignoring_set_id:
+            continue
+
+        # cell_sets_to_return.append(cell_set["key"])
```
ivababukova: remove this comment

python/src/helpers/find_cell_ids_in_same_hierarchy.py
```diff
@@ -0,0 +1,37 @@
...
+        cell_sets_to_return += get_all_cell_ids_in(cell_set)
+
+    return cell_sets_to_return
+
+
+# returns a list with
```
ivababukova: can you finish the sentence in this comment or just remove it?

python/src/helpers/find_cell_ids_in_same_hierarchy.py
```diff
@@ -0,0 +1,37 @@
...
+def find_cell_ids_in_same_hierarchy(key, cell_sets):
+    for cell_set in cell_sets:
+        if cell_set["key"] == key:
+            return get_all_cell_ids_on_current_level(cell_sets, ignoring_set_id=key)
+
+        if cell_set.get("children", None):
+            result = find_cell_ids_in_same_hierarchy(key, cell_set["children"])
+
+            if result:
+                return result
```
ivababukova, line 34: I think you could just `return find_cell_ids_in_same_hierarchy(key, cell_set["children"])` in here?

python/src/tasks/differential_expression.py
```diff
@@ -25,13 +26,39 @@ def _format_result(self, result):
...
+
+        # Ignore basis if it is all cells in the experiment or not specified (default behaviour for previous versions)
+        if not basis_cell_set_name or ("all" in basis_cell_set_name.lower()):
+            return set()
+
+        basis_cell_set = set(find_cells_by_set_id(basis_cell_set_name, resp))
+
+        raw_adata.obs["condition"].loc[
+            ~raw_adata.obs["cell_ids"].isin(basis_cell_set)
+        ] = "FilteredOut"
```
ivababukova, line 42: what does this piece of code do?

cosa65: it is the special string that is used to mark cells not included in the experiment, I'll add a comment explaining that particular part of the function.

python/src/tasks/differential_expression.py
```diff
@@ -42,25 +69,25 @@ def compute(self):
...
                 raw_adata.obs["condition"].loc[
-                    raw_adata.obs["cell_ids"].isin(cells)
-                ] = label
-        
-        raw_adata.obs["condition"].fillna(value="other", inplace=True)
-
-        raw_adata.obs = raw_adata.obs.applymap(lambda x: x.decode() if isinstance(x, bytes) else x)
+                    (raw_adata.obs["cell_ids"].isin(cells))
+                    & (~raw_adata.obs["condition"].eq("FilteredOut"))
+                ] += label
```
marcellp: This does not handle the case that we discussed in Slack.

image

marcellp (changes requested):

python/src/tasks/differential_expression.py
marcellp: I assume the helpers `get_all_cell_ids_on_current_level` and `find_cell_ids_in_same_hierarchy` are there to help you in implementing it.

marcellp (changes requested):
Actually, I see what you are doing here, and it makes sense, but it is very hard to read. I don't think this is a problem with your code, it's more the naming of functions and lack of good documentation.

Could you add some extra comments to the new functions you created as well as the workflow for the `compute()` function in DifferentialExpression so it's easier to understand what the flow is?

cosa65:
> Actually, I see what you are doing here, and it makes sense, but it is very hard to read. I don't think this is a problem with your code, it's more the naming of functions and lack of good documentation.
> 
> Could you add some extra comments to the new functions you created as well as the workflow for the `compute()` function in DifferentialExpression so it's easier to understand what the flow is?

@marcellp 

Looking through this, I think the biggest source of confusion is the 
`for label, name in cell_sets.items():`
cycle, since we are just going through two elements, `first` and `second`.

I will remove it and actually do the corresponding checks for each of these elements, which will make it more readable and easier to modify.

python/src/helpers/find_cell_ids_in_same_hierarchy.py
cosa65: Actually this is not just a recursive call, it is a recursive call on the children cell set, so one level down.

In the case it is not found in the children cell set I need the `for` to iterate over to the next cell set in the same level.
I was trying to follow the logic of `find_cells_by_set_id` that searches in a similar way:

```
        if cell_set.get("children", None):
            result = find_cells_by_set_id(needle, cell_set["children"])

            if result:
                return result
```

ivababukova: I see, yeah you are right.

ivababukova (approved):
lgtm, but we need to work out why the build is failing and fix it before merging

marcellp (approved):

PR #87, opened 2021-03-16
Tests test

ogibson:
# Background

From this PR onwards testing will ocurr with post requests sent from python to the r worker, allowing the r worker to actually create a response. The r response is tested and compared with an expected response for the testing rds object. If we change the rds object then we need to create and curate the corresponding result json files. 

# Changes
### Code changes

# Definition of DONE

### Testing
- [X ] Unit tests written
- [X ] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
- [ X] Photo of a cute animal attached to this PR
image
⇩⇩⇩
This is me when the checks pass.


python/src/tests/tasks/test_cluster_cells.py
```diff
@@ -20,42 +21,50 @@ def load_correct_definition(self):
...
+                "name": "ClusterCells",
+                "cellSetName": "Leiden clusters",
+                "type": "leiden",
+                "cellSetKey": "leiden",
+                "params": {"resolution":0.5},
             },
         }
+        self.correctResponse = json.load(
+            open(os.path.join("tests", "cluster_result.json"))
+        )
```
alexvpickering, line 40: I like the testing strategy of comparing real result to hand currated

python/src/tests/tasks/test_cluster_cells.py
```diff
@@ -20,42 +21,50 @@ def load_correct_definition(self):
...
-
-"""
     def test_louvain_clustering_works(self):
-        res = ClusterCells(self.correct_request, self._adata).compute()
+        res = ClusterCells(self.correct_request).compute()
         res = res[0].result
         res = json.loads(res)
+        """
+        I like these tests but the other one is more powerful.
+        Clustering should be deterministic, let's see if this works fine, if not we can revert to old asserts.
```
alexvpickering: I agree. It might ultimately be usefull to have both to differentiate why something breaks:

1) These very basic tests at least tell us we are getting a response that we expect. 
2) The comparison tests will break as well if the response is malformed but could also break if something benign changes e.g. updates to the underlying clustering code.

python/src/tests/tasks/test_gene_expression.py
```diff
@@ -4,80 +4,73 @@
...
+                "genes": ["Tpt1"],
+            },
+        }
         self.correct_request = {
             "experimentId": "e52b39624588791a7889e39c617f669e",
             "timeout": "2099-12-31 00:00:00",
             "body": {
                 "name": "GeneExpression",
-                "genes": ["PPBP", "PF4"],
+                "genes": ["Tpt1", "Zzz3"],
```
alexvpickering, line 27: Why was this previously asking for human genes (uppercase) from the same experiment?

python/src/tests/tasks/test_list_genes.py
```diff
@@ -8,88 +8,126 @@
...
+                "name": "ListGenes",
+                "selectFields": ["gene_names", "dispersions"],
+                "orderBy": "gene_names",
+                "orderDirection": "DESC",
+                "offset": 0,
+                "limit": 40,
+                "geneNamesFilter": "^Glefjskamsmesascc",
+            }
+        }
+        self.correct_response = json.load(open(os.path.join("tests", "lg_result.json")))
```
alexvpickering, line 76:
```suggestion
        self.correct_response = json.load(open(os.path.join("tests", "list_genes_expected.json")))
```
Might make sense to use explicit file names that match tests (applies to other expected results as well).

python/src/tests/tasks/test_list_genes.py
```diff
@@ -8,88 +8,126 @@
...
     def test_filter_contains_pattern_gets_applied_to_results(self):
-        task_def = self.correct_request_skeleton["body"]
-        task_def["geneNamesFilter"] = "%LIN%"
-        res = ListGenes(self.correct_request_skeleton, self._adata).compute()
+        res = ListGenes(self.correct_filter).compute()
         res = json.loads(res[0].result)
-
         for row in res["rows"]:
-            assert "LIN" in row["gene_names"]
+            assert "lin".casefold() in row["gene_names"].casefold()
```
alexvpickering:
```suggestion
            assert "lin" in row["gene_names"].lower()
```
`casefold()` on hardcoded lowercase string not necessary. Also I've never encountered strange genes name symbols from R that would make casefold necessary over lower.

python/src/tests/tasks/test_list_genes.py
```diff
@@ -8,88 +8,126 @@
...
-        task_def["geneNamesFilter"] = "LIN%"
-        res = ListGenes(self.correct_request_skeleton, self._adata).compute()
+        res = ListGenes(self.correct_startswith).compute()
         res = json.loads(res[0].result)
 
         for row in res["rows"]:
-            assert row["gene_names"].startswith("LIN")
-"""
\ No newline at end of file
+            assert row["gene_names"].casefold().startswith("LIN".casefold())
```
alexvpickering:
```suggestion
            assert row["gene_names"].lower().startswith("lin")
```
Use hardcoded `"lin"` or `"LIN".lower()` for both this and previous test for consistency

alexvpickering (approved):
Looks good to me. I checkout'ed and all tests passed. See comments - all just suggestions/questions though. Good work!

python/src/tests/tasks/test_gene_expression.py
ogibson: It was another experiment!

python/src/tests/tasks/test_cluster_cells.py
ogibson: Good point

PR #93, opened 2021-03-31
Worker to support AWS X-Ray

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-732

https://ui-xray-enable.scp-staging.biomage.net/

https://github.com/biomage-ltd/api/pull/81

# Changes
### Code changes
Adds X-Ray instrumentation to the worker

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

- [x] Relevant wiki pages created/updated

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -1,6 +1,6 @@
 import os
 import types
-from aws_xray_sdk.core import patch_all
+from aws_xray_sdk.core import xray_recorder, patch_all
 
 
 def get_config():
@@ -61,7 +61,8 @@ def get_config():
         config.BOTO_RESOURCE_KWARGS["endpoint_url"] = "http://host.docker.internal:4566"
         config.R_WORKER_URL = "http://r:4000"
 
-    if cluster_env != "development" and cluster_env != "test":
+    if cluster_env != "test":
+        xray_recorder.configure(context_missing='LOG_ERROR')
         patch_all()
 
     return config
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -1,5 +1,6 @@
 import os
 import types
+from aws_xray_sdk.core import patch_all
 
 
 def get_config():
@@ -59,5 +60,8 @@ def get_config():
     if cluster_env == "development":
         config.BOTO_RESOURCE_KWARGS["endpoint_url"] = "http://host.docker.internal:4566"
         config.R_WORKER_URL = "http://r:4000"
+
+    if cluster_env != "development" and cluster_env != "test":
+        patch_all()
 
     return config
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -33,6 +33,7 @@ def get_config():
 
     config = types.SimpleNamespace(
         CLUSTER_ENV=cluster_env,
+        SANDBOX_ID=sandbox_id,
         QUEUE_NAME=queue_name,
         TIMEOUT=timeout,
         IGNORE_TIMEOUT=ignore_timeout,
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -1,7 +1,6 @@
 import os
 import types
-from aws_xray_sdk.core import xray_recorder, patch_all
-from aws_xray_sdk import global_sdk_config
+import aws_xray_sdk as xray
 
 def get_config():
     kube_env = os.getenv("K8S_ENV")
@@ -57,13 +56,13 @@ def get_config():
         config.AWS_ACCOUNT_ID = "000000000000"
         config.BOTO_RESOURCE_KWARGS["aws_access_key_id"] = "my-key"
         config.BOTO_RESOURCE_KWARGS["aws_secret_access_key"] = "my-secret-key"
-        global_sdk_config.set_sdk_enabled(False)
+        xray.global_sdk_config.set_sdk_enabled(False)
 
     if cluster_env == "development":
         config.BOTO_RESOURCE_KWARGS["endpoint_url"] = "http://host.docker.internal:4566"
         config.R_WORKER_URL = "http://r:4000"
 
     if cluster_env != "test":
-        patch_all()
+        xray.core.patch_all()
 
     return config
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -62,7 +62,6 @@ def get_config():
         config.R_WORKER_URL = "http://r:4000"
 
     if cluster_env != "test":
-        xray_recorder.configure(context_missing='LOG_ERROR')
         patch_all()
 
     return config
--- a/python/src/config/__init__.py
+++ b/python/src/config/__init__.py
@@ -1,7 +1,7 @@
 import os
 import types
 from aws_xray_sdk.core import xray_recorder, patch_all
-
+from aws_xray_sdk import global_sdk_config
 
 def get_config():
     kube_env = os.getenv("K8S_ENV")
@@ -57,6 +57,7 @@ def get_config():
         config.AWS_ACCOUNT_ID = "000000000000"
         config.BOTO_RESOURCE_KWARGS["aws_access_key_id"] = "my-key"
         config.BOTO_RESOURCE_KWARGS["aws_secret_access_key"] = "my-secret-key"
+        global_sdk_config.set_sdk_enabled(False)
 
     if cluster_env == "development":
         config.BOTO_RESOURCE_KWARGS["endpoint_url"] = "http://host.docker.internal:4566"
--- a/python/src/consume_message.py
+++ b/python/src/consume_message.py
@@ -1,3 +1,4 @@
+import traceback
 import boto3
 from botocore.exceptions import ClientError
 import json
@@ -43,10 +44,9 @@ def _read_sqs_message():
     try:
         message = message[0]
         print(datetime.datetime.utcnow(), message.body)
-        body = json.loads(message.body)
-        print(datetime.datetime.utcnow(), "Consumed a message from SQS.")
 
-        trace_header = message.attributes.get("AWSTraceHeader", None)
+
+        trace_header = message.attributes and message.attributes.get("AWSTraceHeader", None)
 
         if trace_header:
             global_sdk_config.set_sdk_enabled(True)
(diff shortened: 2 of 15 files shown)


python/src/config/__init__.py
```diff
@@ -1,6 +1,7 @@
 import os
 import types
-
+from aws_xray_sdk.core import xray_recorder, patch_all
+from aws_xray_sdk import global_sdk_config
```
xverges: I would do another kind of import: `patch_all` is too generic a name to not leave you puzzled when you see it
```
import aws_xray_sdk as xray
...
xray.global_sdk_config
...
xray.core.patch_all()
```
or 
```
from aws_xray_sdk import core.patch_all as xray_instrument_services, ..., 
```

python/src/tasks/tasks.py
```diff
@@ -8,6 +8,8 @@
 from .gene_expression import GeneExpression
 from .cluster_cells import ClusterCells
 from result import Result
+from aws_xray_sdk import global_sdk_config
```
xverges: unused

python/src/tasks/tasks.py
```diff
@@ -28,9 +30,9 @@ def submit(self, msg):
         try:
             result = my_class.compute()
             return result
-        except Exception:
+        except Exception as e:
             trace = traceback.format_exc()
-            print(trace)
```
xverges, line 33: Don't we also want the console log?

python/src/response.py
```diff
@@ -46,12 +48,22 @@ def _get_response_msg(self, s3_keys=None):
...
+        was_enabled = global_sdk_config.sdk_enabled()
+
+        if was_enabled:
+            global_sdk_config.set_sdk_enabled(False)
+
         client.put_object(Key=key, Bucket=self.s3_bucket, Body=body)
+
+        if was_enabled:
+            global_sdk_config.set_sdk_enabled(True)
+
```
aerlaut: Can you explain this part Marcell? It seems like the SDK is disabled before putting the object, and then enabled again after that?

marcellp: Correct, the reason is that there is a bug in botocore's S3.

See: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/298

marcellp: Added a comment re this.

python/src/helpers/count_matrix.py
```diff
@@ -22,84 +27,57 @@ def get_objects(self):
...
-
-            with open(file_path, "rb") as f:
-                for chunk in iter(lambda: f.read(partsize), b""):
-                    md5_digests.append(hashlib.md5(chunk).digest())
-
-            candidate = (
-                hashlib.md5(b"".join(md5_digests)).hexdigest()
-                + "-"
-                + str(len(md5_digests))
+            last_mod_local = datetime.datetime.fromtimestamp(
```
xverges, line 42: Could we force the local last modified time to the one in s3? https://docs.python.org/3/library/os.html#os.utime

xverges (commented):
Left some questions/suggestions

python/src/helpers/count_matrix.py
marcellp: This segment is executed only when the worker has been down for a while and is restarted with a file already in there. It is possible that the file we currently have on disk is not the one that is in S3, so we have to compare the last modified of the file with the last modified in S3.

python/src/consume_message.py
```diff
@@ -38,9 +44,29 @@ def _read_sqs_message():
...
+                f"worker-{config.CLUSTER_ENV}-{config.SANDBOX_ID}",
+                traceid=trace_id,
+                sampling=sampled,
+                parent_id=header.parent,
+            )
+    
         body = json.loads(message.body)
         print(datetime.datetime.utcnow(), "Consumed a message from SQS.")
     except Exception as e:
+        xray_recorder.current_segment().add_exception(e, traceback.format_exc())
```
ogibson, line 68: If the xray_recorder.begin excepted will there be a current_segment ?

ogibson: Trying to understand how xray works

python/src/helpers/count_matrix.py
```diff
@@ -22,84 +27,57 @@ def get_objects(self):
...
-
-    def validate_etag(self, file_path, etag):
-
-        # get the size of the file. if it doesn't exist,
-        # the etags definitely don't match
-        try:
-            filesize = os.path.getsize(file_path)
-        except OSError:
-            return False
+    @xray_recorder.capture("CountMatrix.download_object")
```
ogibson, line 34: What information is saved by the recorder when you use capture for a function?

ogibson: I can probably google this for additional info, but just a general pointer on what info will be saved

python/src/helpers/count_matrix.py
```diff
@@ -22,84 +27,57 @@ def get_objects(self):
...
+        if self.last_fetch and last_modified < self.last_fetch:
+            print(
+                datetime.datetime.utcnow(),
+                "Did not fetch as last modified (remote) of",
+                last_modified,
+                "was before last fetch time of",
+                self.last_fetch,
+            )
 
-        print(f"Now checking {key} (etag: {etag}) in S3 against {path}...")
```
ogibson, line 96: So this was taking a lot of time right?

python/src/helpers/count_matrix.py
```diff
@@ -108,10 +86,15 @@ def download_object(self, key, etag):
                 Fileobj=f,
             )
 
+            self.last_fetch = datetime.datetime.now(timezone.utc)
             f.seek(0)
 
+        if was_enabled:
```
ogibson, line 95: Why do this twice? Should the second one set to true?

python/src/config/__init__.py
```diff
@@ -55,9 +57,13 @@ def get_config():
         config.AWS_ACCOUNT_ID = "000000000000"
         config.BOTO_RESOURCE_KWARGS["aws_access_key_id"] = "my-key"
         config.BOTO_RESOURCE_KWARGS["aws_secret_access_key"] = "my-secret-key"
+        global_sdk_config.set_sdk_enabled(False)
```
ogibson: What does sdk_enabled do? (besides enabling sdk) What is sdk?

python/src/helpers/count_matrix.py
```diff
@@ -135,4 +117,7 @@ def sync(self):
             "objects matching experiment.",
         )
 
-        synced = {key: self.download_object(key, etag) for key, etag in objects.items()}
\ No newline at end of file
+        synced = {
+            key: self.download_object(key, last_modified)
```
ogibson, line 124: love that we now use modification date instead of etag

python/src/helpers/count_matrix.py
xverges: I think understand that. My point is that, if we force the local timestamp to be the one of the original file, we are avoiding an improbable race condition where the s3 file has been modified while are downloading it:

1. 07:00 s3 file is modified
2. 08:00 we download s3 file
3. 08:01 s3 file is modified again
4. 08:02 the download is completed and the file is written. If we set its timestamp to now (08:02), we will miss the 08:01 update the next time we run this code. If we set the timestamp to the one that s3 told us about (07:00), we are good for next time.

python/src/helpers/count_matrix.py
```diff
@@ -22,84 +27,57 @@ def get_objects(self):
...
-
-            candidate = (
-                hashlib.md5(b"".join(md5_digests)).hexdigest()
-                + "-"
-                + str(len(md5_digests))
+            last_mod_local = datetime.datetime.fromtimestamp(
+                os.path.getmtime(path), tz=timezone.utc
             )
+        except Exception as e:
+            print(e)
```
ogibson, line 46: should this print be added to current segment? current_segment().add_exception()

python/src/helpers/count_matrix.py
```diff
@@ -22,84 +27,57 @@ def get_objects(self):
...
+        elif last_mod_local and last_modified < last_mod_local:
+            print(
+                datetime.datetime.utcnow(),
+                "Did not fetch as last modified (remote) of",
+                last_modified,
+                "was before last modified (local) of",
+                last_mod_local,
+            )
 
-        if self.path_exists and self.validate_etag(path, etag):
```
ogibson, line 98: So this was taking a lot of time right?

ogibson (commented):
I've left some pretty basic questions to understand how xray works. The code looks good, I'll try it out on local to try an understand a bit more about xray and then approve it asap.

python/src/helpers/count_matrix.py
marcellp: yup

python/src/helpers/count_matrix.py
marcellp: Oh, okay, got it. Let me fix.

python/src/helpers/count_matrix.py
marcellp: No, this will only occur if the file does not exist or there was some kind of error getting the last modified date. We don't need to throw anything here, we can just re-download the file.

python/src/helpers/count_matrix.py
marcellp: It adds an extra event to the trace so we know what function the request was in in that given time. It creates a subsegment with the scope of the function. See, e.g. in the picture, where a subsegment is `WorkSubmitService.getQueueAndHandleMessage` and the AWS call is scoped under it.

image

python/src/consume_message.py
marcellp: No, but then that one will throw an exception and that will be caught by the worker.

python/src/config/__init__.py
marcellp: Enables XRay.

python/src/helpers/count_matrix.py
marcellp: No, `was_enabled` is only checked in line 78. If it was already enabled, we disable and re-enable. If it was never enabled, we don't disable or reenable.

ogibson: Yeah that's what I meant, in the second one you set it to false again instead of re-enabling it

marcellp:
Need to fix the race condition seen [here](https://eu-west-1.console.aws.amazon.com/xray/home?region=eu-west-1#/traces/1-6065f9f5-aa2ce33e623f41e3354d561a) discovered by instrumenting the code.

python/src/helpers/count_matrix.py
marcellp: oh, good catch! fixed :)

ogibson (approved):

PR #107, opened 2021-04-19
Init debug

alexvpickering:
# Background

These changes setup the `DEBUG_STEP` environment variable. This variable is used to save the `req` and `data` objects received by a worker function to allow debugging/developing interactively.

NOTE: the `data` object does not necessarily need to be saved as it never changes (so can be accessed under `data/e52.../r.rds`). I decided to save it regardless with the thought that it will be nice to just click a single object and be able to have everything that goes into a given function.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


r/src/work.r
```diff
@@ -57,6 +61,28 @@ load_data <- function() {
...
+    post_fun(req)
+}
+
+handle_debug <- function(req) {
+    task_name <- basename(req$path)
+    is_debug <- debug_step == task_name
+    
+    if (is_debug) {
+        fname <- paste0(task_name, '.RData')
+        fpath_cont <- file.path('/debug', fname)
```
xverges: Looks like `task_name` will contain slashes that will be interpreted as folder names, won't them?. Is `save()` happy with that?

alexvpickering: Not sure I understand. The `fpath_cont` will be e.g. `'/debug/${task_name}.RData'` which is used as the file path by `save`

xverges: You don't understand because I did not make sense :-D I missed the `basename()` and thought that taskname would be something like `/v0/getEmbedding`. Sorry for the noise!

alexvpickering: Ahhh k np :-D. Always happy to have someone looking out for my bugs :)

README.md
```diff
@@ -118,4 +118,23 @@ extra_hosts:
...
+
+
+
+## Debugging locally
+
+To save the arguments (`req` and `data`) to a worker function, specify DEBUG_STEP. DEBUG_STEP corresponds to the basename of the paths in work.r and will hot-reload if changed at the top of work.r
+
+```bash
+# e.g. DEBUG_STEP=getClusters
+DEBUG_STEP=task_name npm start
```
seb-mueller:
```suggestion
# e.g. DEBUG_STEP=getClusters
DEBUG_STEP=task_name make build && make run
```

seb-mueller: Since we are not using node.js for the worker

seb-mueller (commented):
Really helpful PR!
Just a few comments/questions:
1. `debug/{taskname}.rdata` is overwritten everytime a `taskname` is called, this can be quite taxing since it contains the large seurat object doesn't really change causing a lot of unnessary I/O and might cause hard to debug time-outs. You have pointed this already out, but I think this might become an issues for larger datasets. 
A mitigation might be to save the seurat object seperatly as a file and only update the `req`. The downside being less ease of use (not everything is in one rdata).. I'd still prefer this one seperate option though.
2. If ones changes the experiment, the `./debug/{task}.rdata` object will be overwritten, I'd propose to have experimental ID as part of the filename `./debug/{EID}_{taskname}.rdata` or so.
3. As with pipeline, an `all` option would be very helpful (see below)

A possible strategy for all points could be to save the seurat object as `./debug/{EID}.rds` only if it doesn't exist already (which would save I/O). Alongside an `./debug/{EID}_{taskname}.rdata` only containing the `req` object, this will always be overwritten when taskname is called. The option `all` would than allow to generate multiple rdata objects for each task (which are quick to generate).

README.md
kafkasl: actually now that the Makefile is merged, you can just do `make build && make run`

seb-mueller: Ah, thanks Pol, I've changed it, hope that's fine now (haven't tested it though)

alexvpickering: > Really helpful PR!
> Just a few comments/questions:
> 
> 1. `debug/{taskname}.rdata` is overwritten everytime a `taskname` is called, this can be quite taxing since it contains the large seurat object doesn't really change causing a lot of unnessary I/O and might cause hard to debug time-outs. You have pointed this already out, but I think this might become an issues for larger datasets.
>    A mitigation might be to save the seurat object seperatly as a file and only update the `req`. The downside being less ease of use (not everything is in one rdata).. I'd still prefer this one seperate option though.
> 2. If ones changes the experiment, the `./debug/{task}.rdata` object will be overwritten, I'd propose to have experimental ID as part of the filename `./debug/{EID}_{taskname}.rdata` or so.
> 3. As with pipeline, an `all` option would be very helpful (see below)
> 
> A possible strategy for all points could be to save the seurat object as `./debug/{EID}.rds` only if it doesn't exist already (which would save I/O). Alongside an `./debug/{EID}_{taskname}.rdata` only containing the `req` object, this will always be overwritten when taskname is called. The option `all` would than allow to generate multiple rdata objects for each task (which are quick to generate).

Thanks for the insightfull comments @seb-mueller. I totally agree and will make these changes.

alexvpickering:
> Really helpful PR!
> Just a few comments/questions:
> 
> 1. `debug/{taskname}.rdata` is overwritten everytime a `taskname` is called, this can be quite taxing since it contains the large seurat object doesn't really change causing a lot of unnessary I/O and might cause hard to debug time-outs. You have pointed this already out, but I think this might become an issues for larger datasets.
>    A mitigation might be to save the seurat object seperatly as a file and only update the `req`. The downside being less ease of use (not everything is in one rdata).. I'd still prefer this one seperate option though.
> 2. If ones changes the experiment, the `./debug/{task}.rdata` object will be overwritten, I'd propose to have experimental ID as part of the filename `./debug/{EID}_{taskname}.rdata` or so.
> 3. As with pipeline, an `all` option would be very helpful (see below)
> 
> A possible strategy for all points could be to save the seurat object as `./debug/{EID}.rds` only if it doesn't exist already (which would save I/O). Alongside an `./debug/{EID}_{taskname}.rdata` only containing the `req` object, this will always be overwritten when taskname is called. The option `all` would than allow to generate multiple rdata objects for each task (which are quick to generate).

Thanks for the insightfull comments @seb-mueller. I totally agree and will make these changes.

alexvpickering:
@seb-mueller I made the changes you requested:

1) `data` is copied (faster than saving) only if it doesn't already exist
2) I added `all` as an option for `DEBUG_STEP`
3) `experiment_id` is part of the name for saved `req` and `data` objects.

Thanks again for the great suggestions! Sorry it took me so long to make them :(

docker-compose.linux-dev.yaml
```diff
@@ -1,8 +1,35 @@
 version: "3.2"
```
xverges: @alexvpickering Is there any reason why we cannot use a minimalistic `docker-compose.linux-dev.yaml` with only the overrides/additions to `docker-compose.yaml`, like the one already in https://github.com/biomage-ltd/worker/blob/develop/docker-compose.linux-dev.yaml

This works by running `docker-compose -f docker-compose.yaml -f docker-compose.linux-dev.yaml ...`. If using `make`, the `Makefile` should already been taking care of this.

alexvpickering: Wasn't aware of that functionality - thanks! I'll update

xverges (approved):

PR #109, opened 2021-04-21
Fix for embedding tests and others

ogibson:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-777

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ X] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
- [ X] Photo of a cute animal attached to this PR
image


python/src/tests/tasks/test_embedding.py
```diff
@@ -80,4 +79,3 @@ def test_tsne_edits_object_appropriately(self):
     def test_throws_on_invalid_embedding_type(self):
         with pytest.raises(Exception):
             ComputeEmbedding(self).compute("definitelynotavalidembedding")
```
xverges, line 81: Does not look like we are testing what we say we are testing 🤔 . I think that we need to create something like
```
            "body": {
                "name": "GetEmbedding",
                "type": "definitelynotavalidembedding",
            }
```

ivababukova: Actually, I think in this case the title of the test is wrong, but the actual test contents are correct. If you look at this line: https://github.com/biomage-ltd/worker/blob/master/python/src/tasks/embedding.py#L25 there is no validation of which task type we are doing. Validation of task name happens here: https://github.com/biomage-ltd/worker/blob/master/python/src/tasks/tasks.py#L57 So this test basically checks what will happen if there is no `task_def` defined in `self` of the class.

ivababukova: so I think just renaming the title of the test to should work

ivababukova (approved):
let's merge after you fix the test name

PR #141, opened 2021-06-21
Add ignore.case flag

JuanluOnieva-biomage:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-666?atlOrigin=eyJpIjoiMDQzZTcyZmQwYzg2NGJhY2FhYTI5YzllNTA2NGNhZmYiLCJwIjoiaiJ9

https://ui-juanlu-worker141.scp-staging.biomage.net/

Firstly, I tried to use `grepl`, which has a parameter to flag the `case.ignore`, however, after testing it I found that it not possible to use at the same time the parameter `fixed` (allow us to grep the gene with exactly the same name) and `case.ignore` (for case insensitive).

R code
```
# scdata Seurat object
# dummy: gene query

df <- scdata@misc$gene_annotations
dummy <- "Neb"
df[grepl(dummy, df$name,ignore.case = TRUE),]
df[grepl(dummy, df$name,ignore.case = TRUE, fixed = TRUE),]

genesSubset <- df[grepl(, df$SYMBOL,ignore.case = TRUE),]
```

Output 
```
> df[grepl(dummy, df$name,ignore.case = TRUE),]
## input  name original_name
## ENSMUSG00000053702 Nebl          Nebl
## ENSMUSG00000026950  Neb           Neb
> df[grepl(dummy, df$name,ignore.case = TRUE, fixed = TRUE),]
## input  name original_name
## ENSMUSG00000053702 Nebl          Nebl
## ENSMUSG00000026950  Neb           Neb
## Warning message:
## In grepl(dummy, df$name, ignore.case = TRUE, fixed = TRUE) :
##  argument 'ignore.case = TRUE' will be ignored
```

So I control the case flag by if statement. 

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/r/R/expression.R
+++ b/r/R/expression.R
@@ -12,7 +12,13 @@
 runExpression <- function(req, data) {
     #Get the annotation matrix with the geneid to name translation, and the subset with the correct names.
     df <- data@misc$gene_annotations
-    genesSubset <- subset(df, toupper(df$name) %in% toupper(req$body$genes))
+    ignore.case <- TRUE
+    if (ignore.case){
+        genesSubset <- subset(df, toupper(df$name) %in% toupper(req$body$genes))
+    }else{
+        genesSubset <- subset(df, df$name %in% req$body$genes)
+    }
+    
     #
     #Get the expression values for those genes in the corresponding matrix.
     geneExpression <-data@assays$RNA@data[unique(genesSubset$input),,drop=FALSE]


r/R/expression.R
```diff
@@ -12,7 +12,13 @@
 runExpression <- function(req, data) {
     #Get the annotation matrix with the geneid to name translation, and the subset with the correct names.
     df <- data@misc$gene_annotations
-    genesSubset <- subset(df, toupper(df$name) %in% toupper(req$body$genes))
+    ignore.case <- TRUE
```
robioson, line 15: Why is this set here? Should it be a parameter?

JuanluOnieva-biomage: Yes, it was originally contemplated in this [thread](https://this-is-biomage.slack.com/archives/C01CP46TJ0Z/p1615819973023100). But it has not yet been decided whether it will be implemented or not.

robioson: Could you make it a default argument then, to show that the feature can be used at a later date? i.e. `runExpression <- function(req, data, ignore.case=TRUE)`.

JuanluOnieva-biomage: But, in the case that we add this option in the UI, it should be part of the req. So at the end, the `ignore.case` parameter will belong to `req`.
`req$body$ignore.case`.

alexvpickering: I personally dislike the idea of a checkbox that the user can deselect so that their search becomes case sensitive. To me it adds clutter to the UI without a clear value (why would the user WANT to have to type the case correctly?).

I think the current behaviour is ok (case insensitive) and that added value for this search will come in the form of gene suggestions as you type.

JuanluOnieva-biomage: So do you think that it is better to closed the PR? @alexvpickering

alexvpickering: I think so but maybe I'm not thinking of something. What are you're thoughts?

JuanluOnieva-biomage: Yes, certainly, I don't think there is any use case where it is necessary to switch to sensitive case. Perhaps what might make more sense is to grep by indicating a distance; however that would lead to multiple matches and currently the UI only captures one

PR #164, opened 2021-08-03
fix renv volume issue and switch to umap-learn

alexvpickering:
# Background

https://ui-alex-worker164-pipeline135.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -8,8 +8,7 @@ services:
       args:
         - GITHUB_PAT=${GITHUB_API_TOKEN}
     volumes:
-      - ./r:/src:cached # for watcher
-      - /src/renv/ # distinct renv in host/container
+      - ./r/R:/src/R:cached # for watcher
       - ./data:/data:cached
       - ./data/debug:/debug
     expose:
--- a/r/Dockerfile
+++ b/r/Dockerfile
@@ -26,6 +26,11 @@ RUN R -e "renv::init(bare = TRUE); options(renv.consent = TRUE); renv::restore(l
 COPY renv.lock .
 RUN R -e "options(renv.consent = TRUE); renv::restore(lockfile='renv.lock')"
 
+# install miniconda and python umap-learn for RunUMAP
+ENV RETICULATE_MINICONDA_PATH "/src/renv/python/r-reticulate"
+RUN R -e "reticulate::install_miniconda()"
+RUN R -e "reticulate::conda_install('/src/renv/python/r-reticulate', packages='umap-learn')"
+
 # ---------------------------------------------------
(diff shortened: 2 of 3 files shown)


r/Dockerfile
```diff
@@ -26,6 +26,11 @@ RUN R -e "renv::init(bare = TRUE); options(renv.consent = TRUE); renv::restore(l
 COPY renv.lock .
 RUN R -e "options(renv.consent = TRUE); renv::restore(lockfile='renv.lock')"
 
+# install miniconda and python umap-learn for RunUMAP
+ENV RETICULATE_MINICONDA_PATH "/src/renv/python/r-reticulate"
+RUN R -e "reticulate::install_miniconda()"
+RUN R -e "reticulate::conda_install('/src/renv/python/r-reticulate', packages='umap-learn')"
```
robioson: You should be able to use either `Sys.getenv` or string interpolation here, right?

alexvpickering: Yes you _should_ be able to use `Sys.getenv('RETICULATE_MINICONDA_PATH')`  after it is set. Happy to try if you think it would be better.

robioson: Yes, not that it's crucial but it would be neater to only have the path once imo

docker-compose.yaml
```diff
@@ -8,8 +8,7 @@ services:
       args:
         - GITHUB_PAT=${GITHUB_API_TOKEN}
     volumes:
-      - ./r:/src:cached # for watcher
-      - /src/renv/ # distinct renv in host/container
+      - ./r/R:/src/R:cached # for watcher
```
robioson, line 11: I am not totally sure what these volumes were/are for but I know it works, so I'll approve and maybe you can explain at your leisure

robioson (approved):

ogibson (approved):

docker-compose.yaml
alexvpickering: Basically they are for the watcher so that the worker will restart if the source code changes. The first mounted a volume into the `r/` folder and the second was supposed to prevent the `r/renv/` folder from also being populated from the host. It didn't seem to have the expected behavior so instead I just direct mount into the `r/R` folder where _most_ of the worker source code that we want to watch is.

robioson: Oh, this explains why editing `work.R` didn't trigger a restart when I tried

PR #165, opened 2021-08-04
Use introspection to assign experiments to workers

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-1248

https://ui-marcellp-worker165.scp-staging.biomage.net

N/A

N/A

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [x] Deployed to staging

### Documentation updates

- [x] Relevant wiki pages created/updated

### Approvers

### Just before merging:

### Optional


python/src/worker/config/__init__.py
```diff
@@ -1,24 +1,22 @@
 import os
 import types
+import re
+from logging import info
```
ivababukova: you can remove this import, it is not being used

r/work.R
```diff
@@ -1,5 +1,5 @@
-library(Seurat)
-library(dplyr)
+# library(Seurat)
+# library(dplyr)
```
ivababukova: this needs to be added back

r/work.R
```diff
@@ -159,8 +158,27 @@ create_app <- function(last_modified, data, fpath) {
...
+        experiment_id <- labels["experimentId", ]
+    }
+
+    if(is.na(experiment_id)) {
+        experiment_id <- Sys.getenv("EXPERIMENT_ID", unset = NA)
+    }
+    
+    if(is.na(experiment_id)) {
+        message("No experiment ID label set yet, waiting...")
+        Sys.sleep(5)
```
ivababukova, line 176: why is this in a separate if statements that are the same, rather than under 1 if statement?

ivababukova:
Another thing is also how is this going to run locally? I was looking today with @ogibson-biomage at some PRs that are trying to autoload experiment id locally and I recommended him to pair-up with you on reviewing these, because they look very relevant to what you are doing

r/work.R
marcellp: Essentially, we read from the file. If the file does not contain experiment_id, we read the env variable. If env variable also does not contain it, we're waiting.

r/work.R
marcellp: Thanks, fixed.

python/src/worker/config/__init__.py
marcellp: Fixed.

marcellp:
> Another thing is also how is this going to run locally? I was looking today with @ogibson-biomage at some PRs that are trying to autoload experiment id locally and I recommended him to pair-up with you on reviewing these, because they look very relevant to what you are doing

Using `EXPERIMENT_ID` as an env variable locally will still work, as it is a fallback in case there is nothing in `/etc/podinfo/labels`. It was done like this by design in order to make it work with local development.

r/work.R
ivababukova: indeed, makes sense. Let's keep it this way

ivababukova (approved):

PR #175, opened 2021-08-16
check env to fix local umap

alexvpickering:
neither `umap-learn` nor `uwot` work locally and getting them to work soon is unlikely. Until then, this workaround will at least allow local use of the worker.

**staging:**

https://ui-alex-worker171.scp-staging.biomage.net/


cosa65 (approved):
approved because the worker seems to no longer crash although I still can't get an sns message in the api (which is probably a different issue).

just need to check staging still works before merging

PR #185, opened 2021-08-23
Refactor work loop

marcellp:
# Background
https://github.com/biomage-ltd/api/pull/210

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -196,18 +196,18 @@ jobs:
         env:
           R_IMAGE_NAME: ${{ format('{0}/{1}:{2}-r-ci', steps.login-ecr.outputs.registry, needs.build-docker.outputs.repo-name, needs.build-docker.outputs.tag) }}
 
-      - id: test
-        name: Test, generate code coverage report and upload it
-        run: |-
-          pip install pytest-cov
-          ci_env=`bash <(curl -s https://codecov.io/env)`
+      # - id: test
+      #   name: Test, generate code coverage report and upload it
+      #   run: |-
+      #     pip install pytest-cov
+      #     ci_env=`bash <(curl -s https://codecov.io/env)`
 
-          docker run -v /home/runner/work/worker/worker/data:/data $ci_env -e CI=true --env CLUSTER_ENV=test --net="host" --entrypoint /usr/bin/env $PYTHON_IMAGE_NAME /bin/bash -c "
-            python3 -m pytest --cov=. --cov-report=xml
-            bash <(curl -s https://codecov.io/bash)
-            "
-        env:
-          PYTHON_IMAGE_NAME: ${{ format('{0}/{1}:{2}-python-ci', steps.login-ecr.outputs.registry, needs.build-docker.outputs.repo-name, needs.build-docker.outputs.tag) }}
+      #     docker run -v /home/runner/work/worker/worker/data:/data $ci_env -e CI=true --env CLUSTER_ENV=test --net="host" --entrypoint /usr/bin/env $PYTHON_IMAGE_NAME /bin/bash -c "
+      #       python3 -m pytest --cov=. --cov-report=xml
+      #       bash <(curl -s https://codecov.io/bash)
+      #       "
+      #   env:
+      #     PYTHON_IMAGE_NAME: ${{ format('{0}/{1}:{2}-python-ci', steps.login-ecr.outputs.registry, needs.build-docker.outputs.repo-name, needs.build-docker.outputs.tag) }}
 
       - id: push
         name: Push docker images to ECR
--- a/python/src/worker/config/__init__.py
+++ b/python/src/worker/config/__init__.py
@@ -66,14 +66,14 @@ class Config(types.SimpleNamespace):
             return self.redis
         except AttributeError:
             if cluster_env == "development" or cluster_env == "test":
-                self.redis = redis.Redis({"host": "host.docker.internal", "port": 6379})
+                self.redis = redis.Redis(host="host.docker.internal", port=6379)
             else:
-                self.redis = redis.Redis({
-                    "host": "master.biomage-redis-staging.aykd0e.euw1.cache.amazonaws.com",
-                    "port": 6379,
-                    "ssl": True,
-                    "ssl_cert_reqs": None
-                })
+                self.redis = redis.Redis(
+                    host="master.biomage-redis-staging.aykd0e.euw1.cache.amazonaws.com",
+                    port=6379,
+                    ssl=True,
+                    ssl_cert_reqs=None
+                )
 
             return self.redis
 
--- a/python/src/worker/config/__init__.py
+++ b/python/src/worker/config/__init__.py
@@ -62,19 +62,20 @@ class Config(types.SimpleNamespace):
 
     @property
     def REDIS_CLIENT(self):
-        if self.redis:
+        try:
             return self.redis
-        elif cluster_env == "development" or cluster_env == "test":
-            self.redis = redis.Redis({"host": "host.docker.internal", "port": 6379})
-        else:
-            self.redis = redis.Redis({
-                "host": "master.biomage-redis-staging.aykd0e.euw1.cache.amazonaws.com",
-                "port": 6379,
-                "ssl": True,
-                "ssl_cert_reqs": None
-            })
+        except AttributeError:
+            if cluster_env == "development" or cluster_env == "test":
+                self.redis = redis.Redis({"host": "host.docker.internal", "port": 6379})
+            else:
+                self.redis = redis.Redis({
+                    "host": "master.biomage-redis-staging.aykd0e.euw1.cache.amazonaws.com",
+                    "port": 6379,
+                    "ssl": True,
+                    "ssl_cert_reqs": None
+                })
 
-        return self.redis
+            return self.redis
 
     @property
     def QUEUE_NAME(self):
--- a/python/src/worker/config/__init__.py
+++ b/python/src/worker/config/__init__.py
@@ -56,10 +56,6 @@ class Config(types.SimpleNamespace):
         return self.get_label('experimentId')
 
     @property
-    def SNS_TOPIC(self):
-        return f"work-results-{cluster_env}-{self.get_label('sandboxId', 'default')}"
-
-    @property
     def SANDBOX_ID(self):
         return self.get_label('sandboxId')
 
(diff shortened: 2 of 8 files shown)


python/src/worker/config/__init__.py
```diff
@@ -55,14 +56,27 @@ def get_label(self, label_key, default=None):
...
+    @property
+    def REDIS_CLIENT(self):
+        try:
+            return self.redis
+        except AttributeError:
+            if cluster_env == "development" or cluster_env == "test":
+                self.redis = redis.Redis(host="host.docker.internal", port=6379)
+            else:
+                self.redis = redis.Redis(
+                    host="master.biomage-redis-staging.aykd0e.euw1.cache.amazonaws.com",
```
marcellp, line 72: this needs to be dynamically fetched from aws

python/src/worker/tasks/cluster_cells.py
```diff
@@ -41,7 +43,33 @@ def _format_result(self, raw):
...
+                {
+                    "$prepend": cell_set_object,
+                }
+            ],
+        )
+
+        info(r.status_code)
+    
+    def _format_result(self, cell_set_object):
+        return [Result(json.dumps(cell_set_object))]
```
marcellp: Set `cacheable` to be `False` for clustering to always run (needed for `update_through_api`).

.github/workflows/ci.yaml
```diff
@@ -196,18 +196,18 @@ jobs:
...
+      #   run: |-
+      #     pip install pytest-cov
+      #     ci_env=`bash <(curl -s https://codecov.io/env)`
+
+      #     docker run -v /home/runner/work/worker/worker/data:/data $ci_env -e CI=true --env CLUSTER_ENV=test --net="host" --entrypoint /usr/bin/env $PYTHON_IMAGE_NAME /bin/bash -c "
+      #       python3 -m pytest --cov=. --cov-report=xml
+      #       bash <(curl -s https://codecov.io/bash)
+      #       "
+      #   env:
+      #     PYTHON_IMAGE_NAME: ${{ format('{0}/{1}:{2}-python-ci', steps.login-ecr.outputs.registry, needs.build-docker.outputs.repo-name, needs.build-docker.outputs.tag) }}
```
ivababukova, line 210: we need to enable this again

python/src/worker/response.py
```diff
@@ -61,58 +56,67 @@ def _upload(self, result):
...
+                        'Key': 'experimentId',
+                        'Value': self.request['experimentId']
+                    },
+                    {
+                        'Key': 'requestType',
+                        'Value': self.request['body']['name']
+                    },
+
+                    # TODO: this needs to be removed and a proper
+                    # ACL system implemented before merging.
```
ivababukova: this comment is inaccurate -- we will be implementing a proper ACL system after merging, most likely in the next sprint

python/src/worker/result.py
```diff
@@ -16,18 +16,14 @@ def __init__(
         self.error = error
         self.cacheable = cacheable
 
-    def get_result_object(self, resp_format=False, s3_path=None):
+    def get_result_object(self, resp_format=False):
```
ivababukova, line 19: do we actually use this `resp_format` for anything?

python/src/worker/tasks/__init__.py
```diff
@@ -6,6 +6,8 @@ class Task(ABC):
 
     def __init__(self, msg):
         self.task_def = msg["body"]
+        self.request = msg
```
ivababukova: looks like this `self.request` is only used in the cluster cells task. If that is indeed the case, wouldn't it be better to move this `self.request` to be a variable only for the `ClusterCells` class and add it in its init method?

python/src/worker/response.py
marcellp: fixed

python/src/worker/tasks/__init__.py
marcellp: hmm, i seem to recall there was an issue with it but it seems to be working. i'll put it back there and see

python/src/worker/result.py
marcellp: yes, we use it, when `resp_format` is `True`, the result object is formatted to be uploaded to S3. When it is set to `False`, it is formatted to be sent to the UI (i.e. no inlining is done for the results)

python/src/worker/config/__init__.py
```diff
@@ -88,12 +102,16 @@ def QUEUE_NAME(self):
     LOCAL_DIR=os.path.join(os.pardir, os.pardir, "data"),
 )
 
+config.API_URL=f"http://api-{config.SANDBOX_ID}.api-{config.SANDBOX_ID}.svc.cluster.local:3000"
```
kafkasl, line 105: any reason it's not included in the `Config(...)` constructor?

python/src/worker/response.py
```diff
@@ -61,58 +56,67 @@ def _upload(self, result):
         if was_enabled:
             xray.global_sdk_config.set_sdk_enabled(False)
 
-        client.put_object(Key=key, Bucket=self.s3_bucket, Body=body)
+        client.put_object(Key=ETag, Bucket=self.s3_bucket, Body=body)
```
kafkasl, line 59: aren't ETags susceptible to change for the same object if it's multipart of different regions or stuff like that? is it ok to use them as caching keys? or the results are just retrieved using the tags?

python/src/worker/response.py
```diff
@@ -61,58 +56,67 @@ def _upload(self, result):
...
+
+        client.put_object_tagging(
```
kafkasl, line 61: is this tagging used to find the corresponding files? the name is for for easier debug?

python/src/worker/response.py
```diff
@@ -61,58 +56,67 @@ def _upload(self, result):
...
-            MessageAttributes={
-                "type": {"DataType": "String", "StringValue": "WorkResponse"}
-            },
-            MessageStructure="json",
-        )
-
-        info(f"Request with UUID {self.request['uuid']} successfully sent to SNS, message ID: {r['MessageId']}.")
+            io.Emit(
+                f'{self.request["experimentId"]}-{self.request["body"]["name"]}',
+                self._construct_response_msg(brief=True)
```
kafkasl, line 100: I understand that brief messages don't have their results uploaded into S3, is this stuff like gene lists?

python/src/worker/response.py
```diff
@@ -61,58 +56,67 @@ def _upload(self, result):
...
         if was_enabled:
             xray.global_sdk_config.set_sdk_enabled(True)
 
-        return key
+        return ETag
 
-    def _send_notification(self, mssg):
-        sns = boto3.client("sns", **config.BOTO_RESOURCE_KWARGS)
+    def _send_notification(self):
+        io = Emitter({"client": config.REDIS_CLIENT})
```
kafkasl, line 93: if redis fails then the worker won't be able to notify anything back right? I assume this is ok in the case where the results are uploaded into S3 as the UI will still check if new results are available as a fallback. however, what would happen for brief messages? they'd be simply lost?

python/src/worker/tasks/cluster_cells.py
```diff
@@ -41,7 +44,33 @@ def _format_result(self, raw):
...
             )
-        return [Result(json.dumps(cell_set), cacheable=False)]
+        
+        return cell_set
+    
+    def _update_through_api(self, cell_set_object):
+        r = requests.patch(
+            f"{config.API_URL}/v1/experiments/{self.request['experimentId']}/cellSets",
+            headers={
+                "content-type": "application/boschni-json-merger+json",
```
kafkasl, line 54: are we using this custom content type to be able to merge directly objects (the cell sets) with the json (the stored celll sets) ?

python/src/worker/tasks/cluster_cells.py
```diff
@@ -71,4 +100,9 @@ def compute(self):
         resR = pd.DataFrame(resR)
         resR.set_index("_row", inplace=True)
         resR["cluster"] = pd.Categorical(resR.cluster)
-        return self._format_result(resR)
+
+        # Convert it into a JSON format and patch the API directly
```
kafkasl, line 104: I assume it's actually an object but in JSON format?

python/src/worker/response.py
marcellp: this is an ETag in the traditional sense, not in the S3 sense. it is confusing, but the AWS term is misleading.

ETags are basically hashes that identify whether a certain resource needs to be refetched, see https://en.m.wikipedia.org/wiki/HTTP_ETag

python/src/worker/response.py
marcellp: it's for all of that, but also for setting up custom retention policies for certain types of work (like differential expression) as well as for access control in the future

python/src/worker/response.py
marcellp: brief just means the type of result is the one that is sent to the UI (ie does not have results, only whether or not there was an error and if the result is cacheable)

python/src/worker/tasks/cluster_cells.py
marcellp: we convert python dictionary into json and add it as part of the json patch

python/src/worker/tasks/cluster_cells.py
marcellp: yes, it's a way to update just part of the cell sets, ie the clusters, irrespective of the content the user produced such as custom annotations

python/src/worker/response.py
marcellp: you misunderstood brief, see above

python/src/worker/config/__init__.py
marcellp: yes, you can't access members of config in the initializer of config

.github/workflows/ci.yaml
marcellp: we decided not to enable it and fix it as a separate ticket along with all the other failing worker tests

python/src/worker/config/__init__.py
marcellp: we will fix it as part of a ticket

cosa65 (approved):

python/src/worker/consume_message.py
```diff
@@ -97,11 +109,20 @@ def consume():
 
     if timeout <= datetime.datetime.utcnow():
         info(
-            f"Skipping processing task with uuid {mssg_body['uuid']}"
-            f"{mssg_body['uuid']} as its timeout of {timeout} has expired..."
+            f"Skipping processing task with ETag {mssg_body['ETag']} "
+            f"as its timeout of {timeout} has expired..."
```
ogibson, line 113: Why do we still have timeouts in the worker, if everything we compute will be used?

PR #190, opened 2021-08-30
[BIOMAGE-1376] paginate differential expression results

alexvpickering:
# Background
#### Link to issue 

https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1376

https://ui-oliver-w190.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/python/src/worker/tasks/differential_expression.py
+++ b/python/src/worker/tasks/differential_expression.py
@@ -25,7 +25,6 @@ class DifferentialExpression(Task):
 
         # JSONify result.
         result = json.dumps({"total":total,"rows": result})
-
         # Return a list of formatted results.
         return [Result(result)]
 
@@ -114,12 +113,18 @@ class DifferentialExpression(Task):
         if len(second_cell_set) == 0:
             raise Exception("No cells id fullfills the 2nd cell set.")
 
-        # create request from the corresponding marked cells
         request = {
             "baseCells": [int(x) for x in first_cell_set],
             "backgroundCells": [int(x) for x in second_cell_set],
             "pagination": self.pagination,
         }
+
+        if "filters" in self.pagination:
+            gene_filter = self.pagination["filters"][0]["expression"]
+            regex_chars = "{}|()?¿*+|/.<>"
+            for char in regex_chars:
+                gene_filter = gene_filter.replace(char, "")
+            request["geneNamesFilter"] = gene_filter
 
         # send request to r worker
         r = requests.post(
--- a/python/src/worker/tasks/differential_expression.py
+++ b/python/src/worker/tasks/differential_expression.py
@@ -7,7 +7,9 @@ from aws_xray_sdk.core import xray_recorder
 
 from ..config import config
 from ..helpers.find_cell_ids_in_same_hierarchy import (
-    find_all_cell_ids_in_cell_sets, find_cell_ids_in_same_hierarchy)
+    find_all_cell_ids_in_cell_sets,
+    find_cell_ids_in_same_hierarchy,
+)
 from ..helpers.find_cells_by_set_id import find_cells_by_set_id
 from ..helpers.remove_regex import remove_regex
 from ..helpers.s3 import get_cell_sets
@@ -21,11 +23,11 @@ class DifferentialExpression(Task):
         self.experiment_id = config.EXPERIMENT_ID
         self.pagination = msg["pagination"]
 
-    def _format_result(self, result,total):
+    def _format_result(self, result, total):
         result = result.to_dict(orient="records")
 
         # JSONify result.
-        result = json.dumps({"total":total,"rows": result})
+        result = json.dumps({"total": total, "rows": result})
         # Return a list of formatted results.
         return [Result(result)]
 
@@ -33,7 +35,8 @@ class DifferentialExpression(Task):
     def get_cells_in_set(self, name, resp, first_cell_set_name):
         cells = []
 
-        # If "rest", then get all cells in the same hierarchy as the first cell set that arent part of "first"
+        # If "rest", then get all cells in the same hierarchy as the first cell set
+        #  that arent part of "first"
         if "rest" in name.lower():
             cells = find_cell_ids_in_same_hierarchy(first_cell_set_name, resp)
         else:
@@ -46,9 +49,6 @@ class DifferentialExpression(Task):
         backoff.expo, requests.exceptions.RequestException, max_time=30
     )
     def compute(self):
-        # get the top x number of genes to load:
-        n_genes = self.task_def.get("maxNum", None)
-
         # get cell sets from database
         resp = get_cell_sets(self.experiment_id)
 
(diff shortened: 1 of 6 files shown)


ogibson (approved):

python/src/worker/tasks/differential_expression.py
```diff
@@ -112,12 +113,19 @@ def compute(self):
...
 
-        # create request from the corresponding marked cells
         request = {
             "baseCells": [int(x) for x in first_cell_set],
             "backgroundCells": [int(x) for x in second_cell_set],
+            "pagination": self.pagination,
         }
 
+        if "filters" in self.pagination:
+            gene_filter = self.pagination["filters"][0]["expression"]
```
cosa65, line 119: I would make this a foreach instead of accessing the first element

python/src/worker/tasks/differential_expression.py
```diff
@@ -112,12 +113,19 @@ def compute(self):
...
+            regex_chars = "{}|()?¿*+|/.<>"
```
cosa65: I think we should probably handle all of the filtering logic in one single part, either the R worker or the python worker. I am not sure about this, but do you think it might be a good option to handle all of the filtering in python?

Alternatively, we could handle all of the filtering logic in R, I would send the whole thing (the whole "filters" object) to R and handle all of it there to avoid splitting the management of the filtering between python and R.

r/R/differential_expression.R
```diff
@@ -21,42 +21,46 @@
...
+  # Replace name with Gene names
+  result$gene_names <- data@misc$gene_annotations[
+    match(rownames(result), data@misc$gene_annotations$input), "name"
+  ]
+  result$Gene <- rownames(result)
 
-    # As a first view, order by p_val_adj, to have the most significant at first.
-    result <- result[order(result$p_val_adj, decreasing = F), ]
+  # As a first view, order by p_val_adj, to have the most significant at first.
+  result <- result[order(result$p_val_adj, decreasing = FALSE), ]
```
cosa65, line 45: I think you had mentioned that `F` didn't work, could you make it explicit in a comment to reduce the chance other people end up setting it to `F` again?

r/R/worker_helpers.R
```diff
@@ -97,3 +97,26 @@ getClusters <- function(type, resolution, data) {
   }
   return(data)
 }
+
+handle_pagination <- function(gene_results, offset, limit, order_by, order_decreasing, filter = NULL) {
```
cosa65: Do we need to have `filter` with `NULL` optional parameter? It seems that we are always passing a `filter` param to this function

r/R/worker_helpers.R
```diff
@@ -97,3 +97,26 @@ getClusters <- function(type, resolution, data) {
...
+  if (!is.null(filter)) {
+    gene_results <- gene_results[grepl(filter, gene_results$gene_name, ignore.case = TRUE), ]
+  }
+
+  if(nrow(gene_results)>0){
+    gene_results$full_count <- nrow(gene_results)
+  }else{
+    gene_results$full_count <- list()
```
aerlaut: Shouldn't you just return 0 instead of an empty list? Doesn't `nrow` return a number / integer?

python/src/worker/tasks/differential_expression.py
```diff
@@ -130,9 +138,12 @@ def compute(self):
         result = pandas.DataFrame.from_dict(r.json())
 
         result.dropna(inplace=True)
-
+        total = 0
+        if len(result) > 0:
```
aerlaut: You don't need this if `result["full_count"]` returns an integer

python/src/worker/tasks/differential_expression.py
ivababukova: can you have more than 1 element under filters? And if so, why aren't we looking at all of them? If not, then why is `filters` entry a list, rather than an an object with 1 entry in it?

python/src/worker/tasks/differential_expression.py
ivababukova: Completely agree with Martin here, otherwise we risk duplicating efforts in both python and R and ending up with an unmaintainable code

python/src/worker/tasks/differential_expression.py
```diff
@@ -112,12 +113,19 @@ def compute(self):
...
             "baseCells": [int(x) for x in first_cell_set],
             "backgroundCells": [int(x) for x in second_cell_set],
+            "pagination": self.pagination,
         }
 
+        if "filters" in self.pagination:
+            gene_filter = self.pagination["filters"][0]["expression"]
+            regex_chars = "{}|()?¿*+|/.<>"
+            for char in regex_chars:
+                gene_filter = gene_filter.replace(char, "")
```
ogibson: string.translate()

ogibson: Refactor to a function

python/src/worker/tasks/differential_expression.py
ogibson: Replace for {df: data.frame, total: totalCount}

r/R/differential_expression.R
```diff
@@ -21,42 +21,46 @@
...
-    # Check if the gene_symbol does not appear in annotation. In that case the NA value will be changed to ENSEMBL ID
-    result$gene_names[is.na(result$gene_names)] <- result$Gene[is.na(result$gene_names)]
+  # Check if the gene_symbol does not appear in annotation. In that case the NA value will be changed to ENSEMBL ID
+  result$gene_names[is.na(result$gene_names)] <- result$Gene[is.na(result$gene_names)]
 
+  message("Paginating results:  ", str(result))
+  order_by <- pagination$orderBy
+  order_decreasing <- pagination$orderDirection == "DESC"
+  offset <- pagination$offset
+  limit <- pagination$limit
```
ivababukova, line 57: We need to modify the API schema to make sure we always get these. Or we have a situation when we can break the worker if we don't pass one of these

ivababukova (approved):
Looks good to merge, we can merge, but we mustn't forget about the API schema modifications

r/R/differential_expression.R
ivababukova: I made a ticket for it actually: https://biomage.atlassian.net/browse/BIOMAGE-1420 so that we don't forget

kafkasl (commented):
please use the `python.code-workspace` to develop which has autoformat on save otherwise there are a lot of formatting issues around like missing spaces after commas, unordered imports and more. Using the correct workspace all of that is autoformatted.

I reformatted it already

marcellp (approved):

PR #198, opened 2021-09-14
Fix volcano plot

ogibson:
# Background

https://ui-agi-worker198.scp-staging.biomage.net/

The volcano plot request didn't send any pagination info, and this was breaking the python worker for that task, because it requires pagination. I've modified the python and r worker to return non paginated data (the whole data) if no pagination settings are defined in the request.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [X] Deployed to staging
- [ ] Passed integration tests 

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/python/src/worker/tasks/differential_expression.py
+++ b/python/src/worker/tasks/differential_expression.py
@@ -119,7 +119,7 @@ class DifferentialExpression(Task):
             if "filters" in self.pagination:
                 gene_filter = self.pagination["filters"][0]["expression"]
                 request["geneNamesFilter"] = remove_regex(gene_filter)
-            request["pagination"] = self.pagination,
+            request["pagination"] = self.pagination
 
         # send request to r worker
         r = requests.post(
--- a/python/src/worker/tasks/differential_expression.py
+++ b/python/src/worker/tasks/differential_expression.py
@@ -7,9 +7,7 @@ from aws_xray_sdk.core import xray_recorder
 
 from ..config import config
 from ..helpers.find_cell_ids_in_same_hierarchy import (
-    find_all_cell_ids_in_cell_sets,
-    find_cell_ids_in_same_hierarchy,
-)
+    find_all_cell_ids_in_cell_sets, find_cell_ids_in_same_hierarchy)
 from ..helpers.find_cells_by_set_id import find_cells_by_set_id
 from ..helpers.remove_regex import remove_regex
 from ..helpers.s3 import get_cell_sets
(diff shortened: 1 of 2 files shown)


python/src/worker/tasks/differential_expression.py
```diff
@@ -112,12 +113,13 @@ def compute(self):
         request = {
             "baseCells": [int(x) for x in first_cell_set],
             "backgroundCells": [int(x) for x in second_cell_set],
-            "pagination": self.pagination,
         }
 
-        if "filters" in self.pagination:
-            gene_filter = self.pagination["filters"][0]["expression"]
-            request["geneNamesFilter"] = remove_regex(gene_filter)
+        if self.pagination:
```
ogibson, line 119: {} evaluates to False in python

ivababukova: you don't need this if statement, because in line 25 you are already setting `pagination` to `{}`, so it can never be undefined. Now, if `pagination` is an empty object, then `if "filters" in self.pagination` will just return false.

r/R/differential_expression.R
```diff
@@ -50,17 +49,23 @@ runDE <- function(req, data) {
...
+    if ("geneNamesFilter" %in% names(req$body)) {
+      filter <- req$body$geneNamesFilter
+    }
+
+    result <- handle_pagination(result, offset, limit, order_by, order_decreasing, filter)
 
-  result <- handle_pagination(result, offset, limit, order_by, order_decreasing, filter)
+  }else{
+    result <- list(gene_results = result,full_count=nrow(result))
+  }
```
ivababukova: You can simplify that code if you just swap the logic of the if statements like this:

```
if("pagination" not %in% names(req$body)){
   message("There is no pagination, returning some other awesome results")
   return(list(gene_results = result,full_count=nrow(result)))
}

message("Paginating results:  ", str(result))
pagination <- req$body$pagination
order_by <- pagination$orderBy
order_decreasing <- pagination$orderDirection == "DESC"
offset <- pagination$offset
limit <- pagination$limit

filter <- NULL
if ("geneNamesFilter" %in% names(req$body)) {
  filter <- req$body$geneNamesFilter
}

return(handle_pagination(result, offset, limit, order_by, order_decreasing, filter))
```

python/src/worker/tasks/differential_expression.py
kafkasl: however, if pagination is empty it will not be added to the request. not sure if that's intended or not but it wont' be equivalent. I'm leaving it for now but in a more clear way.

ivababukova: hmm, that's a good point. No idea if intended or not

r/R/differential_expression.R
kafkasl: done

ivababukova (approved):

PR #242, opened 2022-01-26
[BIOMAGE-1672] - Fix pytests

alexvpickering:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1672

https://ui-alex-worker242.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates
- [x] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [x] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [x] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [x] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing

### Documentation updates
- [x] Relevant Github READMEs updated **or** no GitHub README updates required.
- [x] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional


python/src/tests/tasks/test_differential_expression.py
```diff
@@ -53,18 +53,10 @@ def test_throws_on_missing_parameters(self):
...
-        assert isinstance(request, dict)
-
-        # all expected keys are in the request
-        expected_keys = [
-            "baseCells",
-            "backgroundCells",
-        ]
-
-        assert all(key in request for key in expected_keys)
+        with pytest.raises(Exception, match="fullfills the 2nd cell set"):
```
cosa65, line 61: If we are going to test that parameter existence is checked for, we shoudl also have a test that checks for the first parameter existence. 

Since that is not the objective of this ticket, could we add to [the other ticket we created](https://biomage.atlassian.net/browse/BIOMAGE-1694) for this a point on this?

python/src/tests/test_consume_message.py
```diff
@@ -120,19 +121,21 @@ def test_request_with_expired_timeout_is_discarded(self):
...
+    #     with mock.patch("worker.consume_message._read_sqs_message") as m:
+    #         m.return_value = request
+    #         result = consume()
+
+    #         assert result == {
+    #             "experimentId": "random-experiment-id",
+    #             "timeout": "2900-01-01 00:00:00",
+    #             "uuid": "random-uuid",
+    #             "ETag": "random-etag",
+    #         }
```
gerbeldo, line 144: Why is this commented? Maybe add a comment as to why it's still here or just remove it?

python/src/tests/test_response.py
```diff
@@ -13,9 +13,14 @@ class TestResponse:
...
     def load_correct_definition(self):
         self.request = {
+            "body": {
+                "name": "DifferentialExpression",
+            },
             "experimentId": "random-experiment-id",
             "timeout": "2099-12-31 00:00:00",
             "uuid": "random-uuid",
+            "ETag": "random-etag",
+            "socketId": "random-etag"
```
gerbeldo:
```suggestion
            "socketId": "random-socketId"
```

python/src/tests/test_response.py
```diff
@@ -36,222 +41,84 @@ def test_upload_returns_key_with_uuid_as_folder_when_uploading(
...
-
-        resp = Response(self.request, result)
-        result = resp.publish()
-
-        assert result
+        response_msg = resp._construct_response_msg()
+
+        assert response_msg["request"] == self.request
+        assert response_msg["response"]["cacheable"] == True
+
```
gerbeldo: Add assertion for response error (@cosa65 suggestion)

```suggestion
        assert response_msg["request"] == self.request
        assert response_msg["response"]["cacheable"] == True
        assert response_msg["response"]["error"] == False
```

gerbeldo (approved):
we paired up with martin on this, apart from these small nitpicky comments I think it's OK.

python/src/tests/test_consume_message.py
alexvpickering: Because they don't work except locally with inframock. I added a comment to document this. We may want to properly stub in the future so that these test cases can be checked.

python/src/tests/test_response.py
alexvpickering: @cosa65 mentioned that the ETag and socketId have the same value so that's why I did this. Is that correct Martin?

python/src/tests/tasks/test_differential_expression.py
alexvpickering: Sounds good, should I remove this then or just add a point to the other ticket to also add test for first parameter?

python/src/tests/test_response.py
cosa65: I had mentioned one value depended on the other one and they are very similar, but they are different.

It isn't really very important because both values are decided by the UI but just for more information, the `socketId` is usually `"WorkResponse-{Etag}"` except for the case of clustering, where socketId is `"broadcast"`.

python/src/tests/test_response.py
```diff
@@ -36,222 +41,84 @@ def test_upload_returns_key_with_uuid_as_folder_when_uploading(
...
-        resp = Response(self.request, result)
-        result = resp.publish()
-
-        assert result
+        response_msg = resp._construct_response_msg()
+
+        assert response_msg["request"] == self.request
+        assert response_msg["response"]["cacheable"] == True
+        assert response_msg["response"]["error"] == False
+
```
alexvpickering, line 58:
```suggestion
# These tests only work locally with inframock running. Keeping in case
# we want to mock to be able to run these tests.
```

python/src/tests/test_response.py
```diff
@@ -36,222 +41,85 @@ def test_upload_returns_key_with_uuid_as_folder_when_uploading(
...
-
-        assert result
+        response_msg = resp._construct_response_msg()
+
+        assert response_msg["request"] == self.request
+        assert response_msg["response"]["cacheable"] == True
+        assert response_msg["response"]["error"] == False
+# These tests only work locally with inframock running. Keeping in case
+# we want to mock to be able to run these tests.
+    # @mock.patch("boto3.client")
```
alexvpickering:
```suggestion

    # These tests only work locally with inframock running. Keeping in case
    # we want to mock to be able to run these tests.
    # @mock.patch("boto3.client")
```

PR #258, opened 2022-02-15
[BIOMAGE-1723] - New error handling

alexvpickering:


python/src/worker/tasks/differential_expression.py
```diff
@@ -41,7 +41,7 @@ def _format_request(self):
             "baseCells": [int(x) for x in baseCells],
             "backgroundCells": [int(x) for x in backgroundCells],
             "genesOnly": self.task_def.get("genesOnly", False),
-            "comparisonType": self.task_def.get("comparisonType", "within")
+            "comparisonType": self.task_def.get("comparisonType", "within"),
```
ogibson, line 45:
```suggestion
            "comparisonType": self.task_def.get("comparisonType", "within")
```
```suggestion
            "comparisonType": self.task_def.get("comparisonType", "within"),
```

python/src/worker/tasks/factory.py
```diff
@@ -36,7 +35,7 @@ class TaskFactory:
             GetDoubletScore,
             GetMitochondrialContent,
             MarkerHeatmap,
-            GetExpressionCellSets
+            GetExpressionCellSets,
```
ogibson, line 34: I guess this was added by linter?

python/src/worker/tasks/marker_heatmap.py
```diff
@@ -42,52 +42,52 @@ def _format_request(self):
...
+        # raise an exception if an HTTPError if one occurred because otherwise response.json() will fail
+        response.raise_for_status()
+        result = response.json()
+        self.set_error(result)
+        if self.error:
+            return self._format_result(None)
+
+        truncatedExpression = result["truncatedExpression"]
+        rawExpression = result["rawExpression"]
+        work_result = {}
```
ogibson: Why is it called work_result here and result everywhere else? I'd go with the same option everywhere.

ogibson (approved):
I think this looks good! There are a couple of comments, particularly the work_result one needs to be addressed before merging but I haven't found any obvious problems with this, and no errors so far. LGTM

python/src/worker/tasks/background_expressed_genes.py
```diff
@@ -50,15 +50,18 @@ def compute(self):
...
         )
 
-        # raise an exception if an HTTPError if one occurred because otherwise r.json()
+        # raise an exception if an HTTPError if one occurred because otherwise response.json()
         #  will fail
-        r.raise_for_status()
-        r = r.json()
+        response.raise_for_status()
+        result = response.json()
+        self.set_error(result)
```
kafkasl: if there is an `HTTPError` this code will never be reached because an exception will be raised and the code will exit so all this error handling will be ignored. If you look at file `factory.py+63` you will see the exception handling for the compute tasks. This error handling you added, is already there in line 63. 

The only problem is that it currently only uses a generic error message ` "An unexpected error occurred while performing the work."` What you should do here is just raise an exception with the desired error message and modify that method to pick it up, e.g.:
```
error = response.get("error", False)
if error: 
    raise Exception(error)
return self._format_result({"genes": result["genes"]})
```

then in `factory.py`
```
                result = [
                    Result(
                        str(e),
                        error=True,
                    )
                ]
            return res
```

python/src/worker/tasks/background_expressed_genes.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, msg):
 
     def _format_result(self, result):
         # Return a list of formatted results.
-        return Result({"genes": result["genes"]})
+        return Result(result, error=self.error)
```
kafkasl: the error handling should not be here, see below

python/src/worker/tasks/background_expressed_genes.py
```diff
@@ -50,15 +50,18 @@ def compute(self):
...
-        # raise an exception if an HTTPError if one occurred because otherwise r.json()
+        # raise an exception if an HTTPError if one occurred because otherwise response.json()
         #  will fail
-        r.raise_for_status()
-        r = r.json()
+        response.raise_for_status()
+        result = response.json()
+        self.set_error(result)
+        if self.error:
+            return self._format_result(None)
```
kafkasl, line 65: I don't see the benefit of adding the error to the class. Why not just:
```
result = response.json()
error = response.get("error", False)
return self._format_result(result, error)
```
and have format result handle removing the result if there are errors

kafkasl (changes requested):
I am request changes because of:
1. The error handling is already on `factory.py` it should not be passed along to results class. I outlined a possible way of doing it in a comment.
2. If you are refactoring this, maybe we should already take the chance to add "user errors messages" and more detailed messages for us / API. That way we could forward errors to the UI for the user, while also having more detailed logs for us. One way of doing that without changing the `Result` structure would be to have a custom exception:
```
class RException(Exception):
         def __init__(self, user_msg):
             self.user_msg = user_msg
             self.error = error
```
then raise those when there are custom R errors in
```
response.raise_for_status()
result = response.json()

# the error we get from the worker can either be the one for the user or the generic one, or send both
# I would probably get the error for the user, as the worker will have already printed the exception right?
# here I added both for the sake of completeness

usr_msg = result.get("usr_msg", "") 
error = result.get("error", "") 
raise RException(user_msg=usr_msg, error=error)
```

and then only add that information for our custom exceptions:

```
         # default user message
         user_msg = "An unexpected error occurred while performing the work."
         if isinstance(e, RException):
             user_msg = e.user_msg

          result = Result(
                  { 
                            user_msg: user_msg,
                            error_msg: trace
                    },  error=True)
            return result
``` 

This changes would need to be reflected in the schemas to make sure we all agree on a single way of doing that.

python/src/worker/tasks/factory.py
kafkasl: we should add the same linter/formatter as in other projects like utils & inframock. You can see the setup in:
`biomage-utils/biomage-utils.code-workspace`
That file needs the following development dependencies which and config files:
```
dev-requirements.txt
setup.cfg
isort.cfg
```
and they are installed with `make develop` (there are also some other convenient commands like `make fmt` or `make check` but if you use visual studio it will be automatically formatted and checked).

Let me know if you need help setting it up, should be copy-paste.

python/src/worker/tasks/__init__.py
```diff
@@ -14,3 +14,6 @@ def compute(self):
     @abstractmethod
     def _format_result(self, result):
         ...
+
+    def set_error(self, response):
+        self.error = response.get("error", False)
```
kafkasl: this should not be handled here, see comment below

python/src/worker/tasks/marker_heatmap.py
aerlaut: changed to `result`. The previous `result` variable is changed to `response_json`

python/src/worker/tasks/differential_expression.py
aerlaut: The trailing comma is added by linter

aerlaut: You can activate the linter for the python worker by opening the python workspace (i.e. `code python.code-workspace`)

python/src/worker/tasks/factory.py
aerlaut: Yes

aerlaut:
I think it's a good idea to separate worker errors from HTTP errors. In the event of an error, the R worker will already print out the error, so what's more important in this case is the user error to be passed to the UI.

I'll implement your suggestions and push.

python/src/worker/tasks/__init__.py
aerlaut: removed

python/src/worker/tasks/factory.py
ivababukova: Is the linter that we have in the python worker different from the linter in the other projects? Note that to activate the linter in the python worker, you need to open the python worker as a workspace

python/src/worker/tasks/factory.py
```diff
@@ -36,27 +36,53 @@ class TaskFactory:
...
             result = task.compute()
             return result
-        except Exception as e:
-            trace = f"Exception for task {task.__class__.__name__}:\n" \
-                    f"{traceback.format_exc()}"
 
-            error(trace)
-            xray_recorder.current_segment().add_exception(e, trace)
+        except RWorkerException as r:
+            trace = self._log_exception(task)
```
ivababukova: you are not using this `r` anywhere, shouldn't you use it to print what exactly the error is?

python/src/worker/tasks/factory.py
```diff
@@ -36,27 +36,53 @@ class TaskFactory:
...
+                        {
+                            "error": e.message,
+                            "code": e.code,
+                        },
+                        error=True,
+                    )
+                ]
+            return result
+
+        except Exception as e:
```
ivababukova: This is the same piece of code, but for 2 different errors, you can do something like this:

```
        except Exception as e:
            trace = self._log_exception(task)

            # Only send real traces in development.
            if config.CLUSTER_ENV == "development":
                return result = Result(trace, error=True)

            if isinstance(e, RWorkerException):
                return [
                    Result(
                        {
                            "error": e.message,
                            "code": e.code,
                        },
                        error=True,
                    )
                ]

            return [
                    Result(
                        "An unexpected error occurred while performing the work.",
                        error=True,
                    )
                ]
```

python/src/worker/tasks/list_genes.py
```diff
@@ -6,18 +6,22 @@
...
+        total = result["full_count"]
+
         # convert result to list of row dicts
-        result = result.to_dict(orient="records")
+        df = pd.DataFrame.from_dict(result["gene_results"])
+        rows = df.to_dict(orient="records")
 
         # Return a list of formatted results.
-        return Result({"total": total, "rows": result})
+        return Result({"total": total, "rows": rows})
```
ivababukova: I see some additional changes here apart from the error handling. Can you explain what you are doing here and why?

python/src/worker/tasks/background_expressed_genes.py
```diff
@@ -50,15 +51,23 @@ def compute(self):
...
         # send request to r worker
-        r = requests.post(
+        response = requests.post(
             f"{config.R_WORKER_URL}/v0/getBackgroundExpressedGenes",
             headers={"content-type": "application/json"},
             data=json.dumps(request),
         )
 
-        # raise an exception if an HTTPError if one occurred because otherwise r.json()
+        # raise an exception if an HTTPError if one occurred because otherwise response.json()
```
kafkasl: could you fix the comment? remove the extra "if one"  `raise an exception if an HTTPError **if one** occurred`

python/src/worker/tasks/background_expressed_genes.py
aerlaut: moved

python/src/worker/tasks/list_genes.py
aerlaut: The operations here doesn't seem to compute any additional data, just reformatting the outputs. That's why I moved them under the `_format_result` function. What do you think?

python/src/worker/tasks/background_expressed_genes.py
aerlaut: done 👍

python/src/worker/tasks/factory.py
aerlaut: Implemented

python/src/worker/tasks/factory.py
aerlaut: The error message is now used, but this implementation is replaced with the one suggested in https://github.com/hms-dbmi-cellenics/worker/pull/258#discussion_r809837940

python/src/worker/helpers/worker_exception.py
```diff
@@ -0,0 +1,4 @@
+class WorkerException(Exception):
```
kafkasl, line 1: Custom exceptions should be defined package-wide to make sure they are all in the same place. I would suggest to follow [sklearn example ](https://github.com/scikit-learn/scikit-learn/tree/main/sklearn)and place in a `exceptions.py` file at the python worker root level.

python/src/worker/helpers/get_diff_expr_cellsets.py
```diff
@@ -48,11 +49,15 @@ def get_diff_expr_cellsets(
...
-        raise Exception("No cell id fullfills the 1st cell set.")
+        raise WorkerException(
+            "PYTHON_WORKER_INPUT_ERROR", "No cell id fullfills the 1st cell set."
+        )
 
     # Check if the second cell set is empty
     if len(second_cell_set) == 0:
-        raise Exception("No cell id fullfills the 2nd cell set.")
+        raise WorkerException(
+            "PYTHON_WORKER_INPUT_ERROR", "No cell id fullfills the 2nd cell set."
```
kafkasl: I would suggest either
1. Having different exceptions for python & r worker problems, with this approach I think it's easier that someone just adds a different error-code string (and thus we lose the information about whether it's in the R or in python worker)
2. Moving those error codes into a single file and import them from there (I'd place it as the `exceptions.py` file described in next comments)

python/src/worker/helpers/worker_exception.py
kafkasl: Custom exceptions should be defined package-wide to make sure they are all in the same place. I would suggest to follow [sklearn example ](https://github.com/scikit-learn/scikit-learn/tree/main/sklearn)and place in a `exceptions.py` file at the python worker root level.

I suggest having one for each type of worker and add comments about what each field is expected to be. For me it was weird having "text" as error code (but I saw aws does the same) so it'd good to let other people know how those fields should look like and what they are.

python/src/worker/tasks/background_expressed_genes.py
```diff
@@ -50,15 +51,23 @@ def compute(self):
...
+        # raise an exception if an HTTPError occurred
+        # as otherwise response.json() will fail
+        response.raise_for_status()
+        result = response.json()
 
-        return self._format_result(r)
+        error = result.get("error", False)
+        if error:
+            user_message = error.get("user_message", "")
+            err_code = error.get("error_code", "")
```
kafkasl: it's a minor improvement, but considering this is done in a lot of files I would suggest to make the handling easier:

`raiseIfError(result)`

then write the function to be imported by all tasks like:
```
def raiseIfError(result):
    error = result.get("error") 
    if error:
        user_message = error.get("user_message", "Unexpected error happened") 
        err_code = error.get("error_code", MISSING_ERROR_CODE)
        raise WorkerException(err_code, user_message)

```

The default value for `result.get("error") ` is None which is already falsy, no need to add false as default
I would add the default user message. otherwise, the user might get an error box without any text.

python/src/worker/tasks/dotplot.py
```diff
@@ -54,13 +63,23 @@ def compute(self):
...
-        result = r.json()
 
-        return self._format_result(result)
+        # raise an exception if an HTTPError occurred
+        # as otherwise response.json() will fail
+        response.raise_for_status()
+        result = response.json()
+
+        error = result.get("error", False)
+
```
kafkasl: try to be consistent with blank lines and spacing

python/src/worker/tasks/embedding.py
```diff
@@ -27,15 +28,27 @@ def _format_request(self):
...
-        r.raise_for_status()
-        # The index order relies on cells_id in an ascending form. The order is made in the R part.
-        result = r.json()
-        return self._format_result(result)
+        # raise an exception if an HTTPError occurred
+        # as otherwise response.json() will fail
+        response.raise_for_status()
+
+        # The index order relies on cells_id in an ascending form.
+        # The order is made in the R part.
```
kafkasl: why is this comment relevant here? I don't see anything related to it.

python/src/worker/tasks/factory.py
```diff
@@ -36,39 +36,57 @@ class TaskFactory:
...
-            GetExpressionCellSets
+            GetExpressionCellSets,
         )
     }
 
     def __init__(self):
         self.count_matrix = CountMatrix()
         self.count_matrix.sync()
 
+    def _log_exception(self, task, error_object):
```
kafkasl: this function is a great idea, however it's unrelated to the `TaskFactory` and could be reused (and does not need the `self`). I would either place it at the top of the file as a standalone function or move it somewhere it can be reused (maybe helpers?)

python/src/worker/tasks/factory.py
```diff
@@ -36,39 +36,57 @@ class TaskFactory:
...
+            if isinstance(e, WorkerException):
+                return Result(
+                    {
+                        "error_code": e.error_code,
+                        "user_message": e.user_message,
+                    },
+                    error=True,
+                )
+
+                # Only send real traces in development.
```
kafkasl: this comment is out of place

python/src/worker/tasks/factory.py
```diff
@@ -36,39 +36,57 @@ class TaskFactory:
...
-                    f"{traceback.format_exc()}"
 
-            error(trace)
-            xray_recorder.current_segment().add_exception(e, trace)
+        except Exception as e:
+            trace = self._log_exception(task, e)
 
-            # Only send real traces in development.
             if config.CLUSTER_ENV == "development":
                 result = Result(trace, error=True)
```
kafkasl: this would never be returned the way you re-structured the code

python/src/worker/tasks/factory.py
```diff
@@ -36,39 +36,57 @@ class TaskFactory:
...
-            else:
-                result = [
-                    Result(
-                        "An unexpected error occurred while performing the work.",
-                        error=True,
-                    )
-                ]
-            return result
+
+            if isinstance(e, WorkerException):
```
kafkasl: instead of the `isinstance` better to rely on the `except` mechanism (as you already simplified the handling by creating the `self._log_exception(task, e)`. e.g.:
```
except WorkerException as e:
            _ = self._log_exception(task, e)
            return Result(
                    {
                        "error_code": e.error_code,
                        "user_message": e.user_message,
                    },
                    error=True,
                )

except Exception as e:
            trace = self._log_exception(task, e)

            if config.CLUSTER_ENV == "development": # I think we can just remove this as in development we will see the error in the worker very easily
                return Result(trace, error=True)

            return Result(
                {
                    "error_code": "PYTHON_WORKER_GENERIC_ERROR",
                    "user_message": "An unexpected error occurred while performing the work.",
                },
                error=True,
            )
 
```

python/src/worker/tasks/factory.py
```diff
@@ -78,6 +96,4 @@ def _factory(self, msg) -> Task:
         try:
             return self.tasks[task_name](msg)
         except KeyError as e:
-            raise ValueError(
-                f"Task class with name {task_name} was not found"
-            ) from e
+            raise ValueError(f"Task class with name {task_name} was not found") from e
```
kafkasl: changing the exception type in the handling looks weird, it still is a `KeyError` not a `ValueError`. I would either
`raise KeyError(f"Task class with name {task_name} was not found") from e`
or
`raise KeyError(f"Task class with name {task_name} was not found: {e}")`

python/src/worker/tasks/gene_expression.py
```diff
@@ -26,50 +27,54 @@ def _format_request(self):
...
-                # expression = [float(item) for item in view]
-                mean = float(np.nanmean(viewnp))
-                stdev = float(np.nanstd(viewnp))
-                result[gene] = {"truncatedExpression": {}, "rawExpression": {}}
-                result[gene]["rawExpression"] = {
-                    "mean": mean,
-                    "stdev": stdev,
-                    "expression": view,
-                }
+            viewTr = truncatedExpression[gene]
```
kafkasl: please use snake case in python

python/src/worker/tasks/gene_expression.py
```diff
@@ -26,50 +27,54 @@ def _format_request(self):
...
-        if not len(resultR):
-            result[genes[0]] = {
-                "error": 404,
-                "message": "Gene {} not found!".format(genes[0]),
-            }
+        for gene in rawExpression.keys():
 
-        else:
-            for gene in resultR.keys():
+            view = rawExpression[gene]
```
kafkasl: what is a view? is there a more clear name?

python/src/worker/tasks/marker_heatmap.py
```diff
@@ -42,52 +43,59 @@ def _format_request(self):
...
         result = {}
         data = {}
         order = []
-        if not len(resultR):
-            result[genes[0]] = {
-                "error": 404,
-                "message": "Gene {} not found!".format(genes[0]),
+
+        for gene in rawExpression.keys():
+            view = rawExpression[gene]
```
kafkasl: this code is duplicated in the other heatmap move it into a function and call it from both places

r/R/expression.R
```diff
@@ -10,9 +10,21 @@
...
-  return(getExpressionValues(genesSubset, data))
+    #Get the annotation matrix with the geneid to name translation, and the subset with the correct names.
+    df <- data@misc$gene_annotations
+    genesSubset <- subset(df, toupper(df$name) %in% toupper(req$body$genes))
+
+    if (!nrow(genesSubset)) {
+
+        stop(
+            generateErrorMessage(
+                "R_WORKER_GENE_NOT_FOUND",
```
kafkasl: would be nice to have the error codes in a single place (not a must for this PR)

r/R/get_expression_cellsets.R
```diff
@@ -39,7 +49,12 @@ getExpressionCellSetIDs <- function(filters, data) {
 
   # fail if any requested gene names are missing (can't return requested cellset)
   if (anyNA(name_match)) {
-    stop("Requested ExpressionCellset with gene name(s) that are not present.")
+    stop(
+      generateErrorMessage(
+        "R_WORKER_EXPRESSION_NOT_FOUND",
```
kafkasl: better in a single file

r/R/get_metadata_information.R
```diff
@@ -22,7 +22,12 @@ formatMetadataResult <- function(data, column) {
 
   # check if the experiment has specified column
   if (!column %in% colnames(data@meta.data)) {
-    stop(column, " is not computed for this experiment.")
+      stop(
+        generateErrorMessage(
+            "R_WORKER_COLUMN_NOT_FOUND",
```
kafkasl: better in a single file

r/R/marker_heatmap.R
```diff
@@ -14,5 +14,15 @@ runMarkerHeatmap <- function(req, data) {
   top_markers <- getTopMarkerGenes(nFeatures, data, cellSets)
   top_markers <- getMarkerNames(data, top_markers)
 
-  return(getExpressionValues(top_markers, data))
+  res <- getExpressionValues(top_markers, data)
+  if (!length(res$rawExpression)) {
+    stop(
+      generateErrorMessage(
+        "R_WORKER_NO_MARKER_GENES",
```
kafkasl: better in a single file

kafkasl (commented):
I haven't finished reviewing the R part, but I think there are enough comments for now.

python/src/worker/tasks/background_expressed_genes.py
aerlaut: implemented

python/src/worker/helpers/worker_exception.py
aerlaut: With our setup, does that mean that means inside the `src` folder right?

python/src/worker/tasks/embedding.py
aerlaut: I think this must be left over from a previous thing. It's removed now,

python/src/worker/tasks/background_expressed_genes.py
aerlaut: Refactored

python/src/worker/tasks/factory.py
aerlaut: Do we still need this? In development, we can see the stack trace in the worker terminal window, and we don't need to send the trace to the UI in staging or production. I think we can remove this?

aerlaut: I just saw your comment below. I agree that we do not need this anymore, as we can easily see errors in development, so I'm removing it.

python/src/worker/tasks/factory.py
aerlaut: deleted as it is not relevant anymore

python/src/worker/tasks/factory.py
aerlaut: Moved to helpers

python/src/worker/tasks/factory.py
aerlaut: Refactored. Also deleted the logging in development as we can easily see the trace in the terminal window.

python/src/worker/tasks/factory.py
aerlaut: Changed

python/src/worker/helpers/worker_exception.py
kafkasl: yes indeed

python/src/worker/tasks/gene_expression.py
aerlaut: changed

python/src/worker/tasks/gene_expression.py
aerlaut: yea, it should contain the expression of a gene - not all the genes. I've changed the name.

python/src/worker/tasks/marker_heatmap.py
aerlaut: refactored

r/R/expression.R
aerlaut: refactored

r/R/get_expression_cellsets.R
aerlaut: refactored

r/R/get_metadata_information.R
aerlaut: refactored

r/R/errorCodes.R
```diff
@@ -0,0 +1,6 @@
+errorCodes <- list(
```
ivababukova: this same structure in python was called `ErrorCode`, now it is called `errorCodes`. Having sometimes s at the end and sometimes not will be very confusing. I suggest you call them both with `s` or both without `s`. Also, make the R error codes start with capital E for consistency

python/src/worker/helpers/worker_exception.py
aerlaut: moved

r/R/marker_heatmap.R
aerlaut: refactored

python/src/worker/helpers/get_diff_expr_cellsets.py
aerlaut: Implemented

r/R/errorCodes.R
aerlaut: changed

python/src/worker/helpers/process_gene_expression.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+
+
+def process_gene_expression(data):
+
```
kafkasl: no space after function definition

python/src/worker/helpers/process_gene_expression.py
```diff
@@ -0,0 +1,41 @@
...
+    truncated_expr = data["truncatedExpression"]
+    raw_expr = data["rawExpression"]
+
+    result = {}
+
+    for gene in raw_expr.keys():
+
+        raw_gene_expr = raw_expr[gene]
```
kafkasl: no space after iterations start like`for`

python/src/worker/tasks/doublet_score.py
```diff
@@ -18,23 +19,28 @@ def _format_result(self, result):
...
         )
 
-        # raise an exception if an HTTPError if one occurred because otherwise r.json() will fail
-        r.raise_for_status()
+        # raise an exception if an HTTPError occurred
+        # as otherwise response.json() will fail
+        response.raise_for_status()
+
         # The values are ordered by cells id
         # The result contains a list with the doublet scores values
```
kafkasl: remove this comment as the information is not relevant

python/src/worker/tasks/dotplot.py
```diff
@@ -54,13 +63,17 @@ def compute(self):
...
-        result = r.json()
 
-        return self._format_result(result)
+        # raise an exception if an HTTPError occurred
+        # as otherwise response.json() will fail
+        response.raise_for_status()
+        result = response.json()
+        raise_if_error(result)
+
+        data = result.get("data")
```
kafkasl, line 76: try to be consistent with blank lines, in other files you left one after this

kafkasl (approved):
Great PR, thanks for amending all the issues. The code is fantastic and looks very concise. Keep up the good work

python/src/tests/tasks/test_cluster_cells.py
```diff
@@ -44,3 +46,24 @@ def test_format_request(self):
...
+        payload = {"error": {"error_code": error_code, "user_message": user_message}}
+
+        responses.add(
+            responses.POST,
+            f"{config.R_WORKER_URL}/v0/getClusters",
+            json=payload,
+            status=200,
+        )
+
+        with pytest.raises(RWorkerException) as exc_info:
```
aerlaut: rename `exc_info` to `exception_info`

r/R/marker_heatmap.R
```diff
@@ -14,5 +14,15 @@ runMarkerHeatmap <- function(req, data) {
...
-  return(getExpressionValues(top_markers, data))
+  res <- getExpressionValues(top_markers, data)
+  if (!length(res$rawExpression)) {
+    stop(
+      generateErrorMessage(
+        error_codes$NO_MARKER_GENES,
+        "Couldn't get marker genes"
+      )
+    )
+  }
```
alexvpickering: @ogibson-biomage do we need this? This was a check that was moved from the python code to R code but I can't imagine a scenario where we actually get here (making it hard to test/pacify codecov). I think it was added there because the python code was duplicated from another task but let me know if I'm wrong

python/src/exceptions.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Raises:
+        RWorkerException: Raise R worker exception containing the error code
+            and user message of the error
+    """
+    error = result.get("error")
+    if error:
+        user_message = error.get("user_message", "Unexpected error occured")
+        err_code = error.get("error_code", ErrorCodes.R_WORKER_ERROR)
+        raise RWorkerException(err_code, user_message)
```
alexvpickering: changing since R worker will always return a code and message

alexvpickering:
LGTM. I made most of the changes we discussed today (please check) and added tests for all the added R code. Don't forget to look into the error you found in the python error handling

aerlaut:
> LGTM. I made most of the changes we discussed today (please check) and added tests for all the added R code. Don't forget to look into the error you found in the python error handling

Fixed in [9e1842b](https://github.com/hms-dbmi-cellenics/worker/pull/258/commits/9e1842bd6c315ba03e9892faa132e861e9021fea). Issue was because subclassing `Enum` returns a `key:value` construct instead of the actual value when accessing the variable, and this causes an error when the class is being stringified in preparation for response. Not subclassing from `Enum` returns the actual value, so the error can be stringified.

r/R/marker_heatmap.R
ogibson: I've never seen this stop actually executed. Fell free to remove this check

aerlaut: removed

PR #264, opened 2022-03-02
Move expression code to r

ogibson:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1731

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

- [x] Started end-to-end tests on the latest commit.

### Documentation updates

### Optional


r/tests/testthat/test-expression.R
```diff
@@ -13,27 +13,54 @@ mock_scdata <- function() {
...
-  expect_equal(ncol(res$rawExpression), ncol(res$truncatedExpression))
 
-  expect_equal(nrow(res$rawExpression), ncol(data))
-  expect_equal(ncol(res$rawExpression), 2)
+  expect_equal(length(raw_expression), length(truncated_expression))
+
+  expect_equal(length(raw_expression), ncol(data))
 })
 
+test_that("Expression result is properly formated.", {
```
gerbeldo:
```suggestion
test_that("Expression result is properly formatted.", {
```

r/tests/testthat/test-expression.R
```diff
@@ -13,27 +13,54 @@ mock_scdata <- function() {
...
+
+  expect_equal(raw_expression_result$mean, mean(raw_expression_result$expression, na.rm = TRUE))
+  expect_equal(raw_expression_result$stdev, sd(raw_expression_result$expression, na.rm = TRUE))
+  expect_equal(truncated_expression_result$min, min(truncated_expression_result$expression, na.rm = TRUE))
+  expect_equal(truncated_expression_result$max, max(truncated_expression_result$expression, na.rm = TRUE))
+})
+
+
+
 test_that("Expression task continues if gene doesnt exist and returns appropiate number of results", {
```
gerbeldo:
```suggestion
test_that("Expression task continues if gene doesn't exist and returns appropriate number of results", {
```

r/tests/testthat/test-expression.R
```diff
@@ -13,27 +13,54 @@ mock_scdata <- function() {
...
   data <- mock_scdata()
   req <- mock_req()
 
   res <- runExpression(req, data)
+
   expect_equal(length(res), 2)
+  expect_equal(as.list(names(res)), req$body$genes)
+})
+
+test_that("Expression task returns appropiate number of cells.", {
```
gerbeldo:
```suggestion
test_that("Expression task returns appropriate number of cells.", {
```

r/tests/testthat/test-expression.R
```diff
@@ -13,27 +13,54 @@ mock_scdata <- function() {
   return(pbmc_small)
 }
 
-test_that("Expression task returns appropiate number of cells and genes.", {
+test_that("Expression task returns appropiate number and names of genes.", {
```
gerbeldo:
```suggestion
test_that("Expression task returns appropriate number and names of genes.", {
```

r/tests/testthat/test-embedding.R
```diff
@@ -0,0 +1,37 @@
...
+test_that("PCA embedding works", {
+  data <- suppressWarnings(mock_scdata())
+  req <- mock_req()
+  req$body$type <- "pca"
+
+  res <- runEmbedding(req, data)
+
+  expected_res <- as.data.frame(Seurat::Embeddings(data)[,1:2])
+
+  expected_res <- expected_res %>% as.data.frame %>% rowwise %>% mutate(PCS = list(c(PC_1,PC_2)))
```
gerbeldo:
```suggestion
  expected_res <- expected_res %>%
    as.data.frame() %>%
    dplyr::rowwise() %>%
    dplyr::mutate(PCS = list(c(PC_1, PC_2)))
```

r/R/embedding.R
```diff
@@ -1,79 +1,85 @@
...
-                        min.dist = config$minimumDistance,
-                        metric = config$distanceMetric,
-                        umap.method = "umap-learn")
+  # Order embedding by cells id in ascending form
+  df_embedding <- as.data.frame(df_embedding)
+  df_embedding$cells_id <- data@meta.data$cells_id
+  df_embedding <- df_embedding[order(df_embedding$cells_id), ]
+  df_embedding <- df_embedding %>%
+    tidyr::complete(cells_id = seq(0, max(data@meta.data$cells_id))) %>%
+    select(-cells_id)
```
gerbeldo:
```suggestion
    dplyr::select(-cells_id)
```

r/R/utilities_expression.R
```diff
@@ -0,0 +1,62 @@
...
+  colnames(geneExpression) <- genes$name[symbol_idx]
+  adjGeneExpression <- as.data.frame(apply(geneExpression, 2, FUN = function(x) {
+    lim <- as.numeric(quantile(x, quantile_threshold, na.rm = TRUE))
+    i <- 0.01
+    while (lim == 0 & i + quantile_threshold <= 1) {
+      lim <- as.numeric(quantile(x, quantile_threshold + i, na.rm = TRUE))
+      i <- i + 0.01
+    }
+    return(pmin(x, lim))
+  }))
```
gerbeldo, line 32: I would extract this function to apply. It's a bit too big for an anonymous...

gerbeldo (approved):
looks good. couple of small comments

r/R/embedding.R
```diff
@@ -1,79 +1,85 @@
...
+# runEmbedding
+# Returns a list of x,y coordinates for each cell.
+# req is the request.
 #
-#Req$body has:
-#type = type of embedding, supported umap, pca and tsne.
-#config = config list.
+# Req$body has:
+# type = type of embedding, supported umap, pca and tsne.
+# config = config list.
```
ivababukova, line 7: Why wasn't this formatted well in the first place? Is everyone using the same linter for R?

r/man/load_experiment_data.Rd
```diff
@@ -1,5 +1,5 @@
 % Generated by roxygen2: do not edit by hand
-% Please edit documentation in R/development_helpers.R
```
ivababukova, line 2: what are these files?

ivababukova: where do they get generated from?

r/R/utilities_expression.R
```diff
@@ -0,0 +1,62 @@
+#' Returns expression values for selected genes
+#'
+#' @param genes - Must have names and input(ensmbl ids)
+#' @param data
+#'
+#' @return
+#' @export
+#'
+#' @examples
+getExpressionValues <- function(genes, data) {
```
ivababukova, line 10: instead of having `utilities_x` (like `utilities_development`, `utilities_expression`, etc), why don't you have a folder called "utilities" and inside that folder have "expression.R", "development.R", etc?

ivababukova: Also, utilities is a pain to spell, instead of it I would use something easier to spell as a name, for example "utils"

ivababukova:
Excuse me if I am missing out something obvious, but wasn't the goal of this ticket supposed to be about moving some python code from the expression, marker_heatmap and list_genes tasks to R? At least that's what the title of this ticket: https://biomage.atlassian.net/browse/BIOMAGE-1731 says. All I can see here is only R code being moved around. Are you planning to move the python code as a next iteration after this?

r/R/utilities_expression.R
ivababukova: to be consistent with the python worker, I would actually name the folder not `utilities` or `utils`. I would name it `helpers`.
And in general, I recommend trying to define a structure for those files under `R`, or better follow the structure of the python worker. With lack of structure, it will get very messy when we have even more files (for an example of what happens when not structuring things, look at the utils folder in the UI)

r/R/utilities_development.R
```diff
@@ -0,0 +1,11 @@
+#' Loads Experiment data
+#'
+#' @param exp_id
+#'
+#' @return
+#' @export
+#'
+#' @examples
+load_experiment_data <- function(exp_id = "test") {
+  return(readRDS(paste0("../data/", exp_id, "/r.rds", sep = "")))
```
ivababukova, line 10: don't you have in R an equivalent of `os.path` in python that you can use here? Also, there is a second data folder under `python/src/tests`, I recommend that you either rename the data folder in python or move the contents of that data folder in python to be under the `worker/data` folder to avoid ambiguity

ivababukova, line 10: in this `worker/data` folder, under `test` I can only see `r.rds.gz`, not `r.rds`. How does it pick the `r.rds.gz` file then?

ivababukova, line 10: instead of having to manually insert `/` after each folder name, I recommend that you change `sep` to be `/`

r/R/utilities.R
```diff
@@ -0,0 +1,29 @@
...
+# subset_ids subsets a seurat object with the cell ids
+#
+subset_ids <- function(scdata, cells_id) {
+  meta_data_subset <- scdata@meta.data[match(cells_id, scdata@meta.data$cells_id), ]
+  current_cells <- rownames(meta_data_subset)
+  scdata <- subset(scdata, cells = current_cells)
+  return(scdata)
+}
+
+insertSetChildThroughApi <- function(new_cell_set, api_url, experiment_id, cell_set_key, auth_JWT) {
```
ivababukova: I see two different naming conventions used here for naming functions - `insertSetChildThroughApi` with camel case and `subset_ids` with underscore, like you'd write python code. I don't know what is the standard in R. If there are no standards in R, I suggest you decide on which style you like and then you use on it consistently across the code. Same applies for variable names and also file names.

ivababukova: `insertSetChildThroughApi` seems to be about patching the cell sets, adding a new cluster to it.

This is subjective: for me, it is not obvious from that function name, which makes the code less readable. I would rename the function to show that we are saving a new cell set - like "persistCellset" or `saveCellsetToApi`, `sendCellsetToApi` or something like that.

ivababukova: instead of passing both `experiment_id` and `api_url` as parameters, you can import config in this file and directly use them from the imported config.

r/R/utilities_development.R
```diff
@@ -0,0 +1,11 @@
+#' Loads Experiment data
+#'
+#' @param exp_id
+#'
+#' @return
+#' @export
+#'
+#' @examples
+load_experiment_data <- function(exp_id = "test") {
```
ivababukova, line 9: that looks like a function for setting up testing, I would move the file to be under `tests/`

r/R/utilities_expression.R
```diff
@@ -0,0 +1,62 @@
...
+  quantile_threshold <- 0.95
+  #
+  # Get the expression values for those genes in the corresponding matrix.
+  geneExpression <- data@assays$RNA@data[unique(genes$input), , drop = FALSE]
```
ivababukova, line 14: why do we have this `,  ,` empty space between commas here? It looks easy to omit and delete in the future. Is there some parameter that we can set there to explicitly indicate that we don't want whatever we don't want to happen?

r/R/utilities_expression.R
```diff
@@ -0,0 +1,62 @@
...
+  geneExpression <- as.data.frame(t(as.matrix(geneExpression)))
+  geneExpression$cells_id <- data@meta.data$cells_id
+  geneExpression <- geneExpression[order(geneExpression$cells_id), ]
+  geneExpression <- geneExpression %>%
+    tidyr::complete(cells_id = seq(0, max(data@meta.data$cells_id))) %>%
+    dplyr::select(-cells_id) %>%
+    as.data.frame()
+  # worried about duplicate gene row.names in @data
```
ivababukova: what do you want to say with this comment?

r/R/utilities_expression.R
```diff
@@ -0,0 +1,62 @@
...
+  symbol_idx <- match(colnames(geneExpression), genes$input)
+  colnames(geneExpression) <- genes$name[symbol_idx]
+  adjGeneExpression <- as.data.frame(apply(geneExpression, 2, FUN = function(x) {
```
ivababukova, line 24: I see you are trying to adjust Gene expression in some way. Can you explain what are you adjusting in particular?

r/tests/testthat/test-embedding.R
```diff
@@ -0,0 +1,37 @@
...
+    row.names = paste0("ENSG", seq_len(nrow(pbmc_small)))
+  )
+
+  pbmc_small <- Seurat::RunPCA(pbmc_small, npcs = 5)
+  pbmc_small$samples <- rep("Sample1", 80)
+  pbmc_small@misc$numPCs <- 5
+  return(pbmc_small)
+}
+
+test_that("PCA embedding works", {
```
ivababukova, line 25: are these really the only test cases for embedding? `embedding.R` is 85 lines long, there must be other cases to test for

ivababukova: some example cases to test for:

- when there is `numPCs` in `data@misc` and when there is no `numPCs`
- when there is `active.reduction` in `data@misc` and when there is no `active.reduction`
- when there are some wrong values for `active.reduction` and `numPCs`
- when there is no `misc` in `data`
- this line 44 in `embedding.R`: `df_embedding <- Embeddings(data, reduction = type)[, 1:2]` the `[, 1:2]` bit looks very hacky, you can test what happens if `Embeddings` returns something else. You can also go back to the code and try to make it more durable.
- when `type` is not defined or none
- is the data ordered by cell ids in ascending form (I am looking at lines 68-74 in `embedding.R`)
- what if there were no cells in the data? How will lines 68-74 in `embedding.R` behave?

r/tests/testthat/test-embedding.R
```diff
@@ -0,0 +1,37 @@
...
+}
+
+mock_scdata <- function() {
+  data("pbmc_small", package = "SeuratObject", envir = environment())
+  pbmc_small$cells_id <- 0:(ncol(pbmc_small) - 1)
+  pbmc_small@misc$gene_annotations <- data.frame(
+    input = paste0("ENSG", seq_len(nrow(pbmc_small))),
+    name = row.names(pbmc_small),
+    row.names = paste0("ENSG", seq_len(nrow(pbmc_small)))
+  )
```
ivababukova, line 17: I see that this `mock_scdata` is used in all test functions. You can extract it in a different place and then import and reuse it in individual tests to reduce the copy-paste

ivababukova: also, you might want to consider adding now or in the future other types of datasets for testing with specific characteristics to test for wider variety of cases.

r/tests/testthat/test-expression.R
```diff
@@ -13,27 +13,54 @@ mock_scdata <- function() {
...
   data <- mock_scdata()
   req <- mock_req()
   req$body$genes <- append(req$body$genes, "aaa")
 
   res <- runExpression(req, data)
-  expect_equal(ncol(res$rawExpression), 2)
+  expect_equal(length(res), 2)
 })
 
 test_that("Expression task works with one gene", {
```
ivababukova, line 66: what about a test whether it works with multiple genes?

ivababukova: what about a test to check if the expression task is case sensitive? We had this problem a few months ago, where gene search was failing unless the user types in the gene using exact same letter cases.

r/tests/testthat/test-expression.R
```diff
@@ -42,8 +69,19 @@ test_that("Expression task works with one gene", {
...
+  req <- mock_req()
+
+  data@assays$RNA@data["MS4A1",] <- 0
+  data@assays$RNA@data["MS4A1",1] <- 5
+
+  res <- runExpression(req, data)
+  expect_false(all(res$MS4A1$truncatedExpression$expression==0) )
 })
 
 test_that("runExpression throws an error if request only non existing genes", {
```
ivababukova, line 87: what happens if the user searches for, say 3 genes: one that doesn't exist and two that exist? Will the code still throw an error and not return all of them? Or will it return only the genes that it have?

r/tests/testthat/test-marker-heatmap.R
```diff
@@ -26,14 +26,14 @@ test_that("Marker Heatmap returns appropiate format", {
 
   res <- runMarkerHeatmap(req, data)
 
-  expect_equal(names(res), c("rawExpression", "truncatedExpression"))
+  expect_equal(names(res[[1]]), c("rawExpression", "truncatedExpression"))
 
   # number of rows is number of cells
-  expect_equal(ncol(data), nrow(res$rawExpression))
+  expect_equal(ncol(data), length(res[[1]]$rawExpression$expression))
```
ivababukova, line 32: is there a way to avoid this `res[[1]]`? Looks very fragile. I saw it in other places too. What is `res[[1]]`? What is inside `res`?

r/R/utilities.R
ogibson: subset_ids is wrong. I will change that. We already agreed that R functions will be camel case

r/R/utilities_development.R
ogibson: This file should be deleted. This was an old  set of development functions that I used locally and shared in the repo for other bioinformaticians.

r/R/utilities_expression.R
ogibson: Best coding practices in R indicate the usage of a one level folder named R where the code files are set. We based this structure in some major R packages. For example:[ dplyr](https://github.com/tidyverse/dplyr/tree/main/R).

Dplyr itself uses "utils", I based this on another package that used "utilities". I think this is a good idea, I will rename

r/R/utilities_expression.R
ogibson: It's common R syntax. You wouldn't delete that if you know what you are doing because then you are not selecting columns from that data frame

r/R/utilities_expression.R
ogibson: I think it's a remnant from the past.

r/R/utilities_expression.R
ogibson: This is used to generate the truncated gene expression, where we truncate values to a 95th percentile in order to improve visualization

r/tests/testthat/test-embedding.R
ogibson: Each dataset is different depending on the set of tests, that's why we have a different one per file. I'm happy to chat about changing this if you think it should be changed

r/tests/testthat/test-embedding.R
ogibson: Embedding tests are missing because we were waiting on some different issues. I will create a ticket to add what we can, but embedding can't be reproduced in the worker, so it's difficult to test bc the result will vary.

r/R/utilities_expression.R
ivababukova: can you remove it in this case? Or rephrase it.

ogibson:
```suggestion
```

ogibson: Yes, I've pushed the commit :)

r/tests/testthat/test-embedding.R
ivababukova: I didn't realise this, makes sense! Then leave it as is

r/R/utilities_expression.R
ivababukova: Interesting, I didn't know that. Do you know the reason why this is the recommended way?

r/tests/testthat/test-marker-heatmap.R
ogibson: Res is a list, and we are accessing the first element in the list

r/R/embedding.R
ogibson: We now all use the same linter, which is the Rstudio linter so this shouldn't happen. We have a ticket to add automatic linting to PRs

r/man/load_experiment_data.Rd
ogibson: These are documentation files from roxygen, they are the documentation of our python package, explaining what each function does, etc.
Then when in Rstudio you do ?function you get help!

r/tests/testthat/test-expression.R
ogibson: Addressed in a new ticket!

r/tests/testthat/test-expression.R
ogibson: Addressed in a new ticket!

PR #269, opened 2022-04-18
[BIOMAGE-1865] - bump Seurat to 4.0.6

alexvpickering:


alexvpickering:
setting the seed also didn't help (a la `set.seed(0)`)

PR #333, opened 2023-08-23
added better worker messages

kafkasl:


python/src/worker/helpers/send_status_updates.py
```diff
@@ -0,0 +1,52 @@
...
+        [c if not c.isupper() else f" {c}" for c in name_without_get]
+    ).strip()
+
+    return final_name
+
+
+def format_user_message(status_code, request):
+    task_name = format_task_name(request)
+    if status_code == DOWNLOAD_EXPERIMENT or status_code == LOAD_EXPERIMENT:
+        return "Accessing the Seurat object for your analysis"
```
StefanBabukov, line 32: You can also return those codes from here to the UI instead of having them there too. That way when you receive an error in the UI, you will understand where it comes from if its something like `WORKER_DOWNLOAD_EXPERIMENT`.

kafkasl: yes this is done already, this function only formats the name. See this below:
```
            "type": "WorkResponse",
            "status_code": status_code,
            "user_message": format_user_message(status_code, request),
```

saracastel (approved):

PR #335, opened 2023-09-01
Worker submit performance improvements

cosa65:
# Description

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/17

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


python/src/worker/response.py
```diff
@@ -30,25 +32,47 @@ def _construct_data_for_upload(self):
                 zipfile.write(self.result.data)
             else:
                 info('Encoding and compressing json work result')
-                json.dump(self.result.data, zipfile)
+                ujson.dump(self.result.data, zipfile)
 
 
+
+        gz_body_bytes = None
+        # If size is less than 2 mb, then send it over notification too
```
cosa65: "Send it directly over socket"*

python/src/worker/response.py
```diff
@@ -30,25 +32,47 @@ def _construct_data_for_upload(self):
...
+        # Needs to be done here because upload_fileobj closes the file:
+        # https://github.com/boto/boto3/issues/929
+        mb = 1000000
+        body_size = sys.getsizeof(gzipped_body)
+        info(f"Body size is {body_size}")
+        if (body_size <= 2 * mb):
```
cosa65: I still need to test an fine-tune this cutoff a bit to make sure we switch to sending the `signedUrl` before the data becomes too big for the socket (sockets are not fast for transmitting large pieces of data)

kafkasl: I am not very much in favor of these kind of fine-tunings. It looks like they depend too much on network speed so how do you plan to find an "optimal" value?

python/src/worker/response.py
kafkasl: I think we should have a design doc discussing this. It greatly changes the current flow of data and I'd like to see if it's worth the increased complexity of the resulting code.

python/src/worker/response.py
kafkasl: interesting and relevant answer in stack overflow https://stackoverflow.com/a/12978658.

I think the things to keep in mind are:
1. would all the small request be sent sequentially? 
2. is sending through socket blocking? 
3. is sending to socket faster / slower (from the executing code point of view) than upload to S3?
4. would the message be sent to multiple UIs if you have the experiment opened in many tabs?

cosa65: Thanks for the link! It definitely brings up some interesting points. I would say to keep in mind though that we are not using just sockets, but rather a wrapper for redis's pub/sub mechanism, you can check that we use a library called `socket_io_emitter`. My understanding is that it does end up as a socketio message when reaching the ui though.

I definitely think 2mb is way too much, I had just set a big value to begin the comparison on. The actual value we'll settle on will probably be way smaller.

> I am not very much in favor of these kind of fine-tunings. It looks like they depend too much on network speed so how do you plan to find an "optimal" value?

I am saying fine-tuning, but not going overboard with this at all. I am performing some tests both from my place (SA) and am planning to get help from someone else in europe too to compare whether the gain is still worth it from a place closer to the server itself. I'll update here with the stats I pick up.

1. I didn't understand this question
2. It isn't blocking. I believe there is a blocking emit which cna be used, but we are not.
3. From the informal tests I had done originally, which prompted me to implement this, it does make a difference on small enough messages (in average, from south america, it makes 1 sec of difference). The question of approximately how small the message should be is what I was planning to use some stats for.
4. No, only if all tabs are performing the same work request at the same time. The only messages that arrive to all open tabs of an experiment in particualr are the ones sent to the `ExperimentUpdates-{experimentId}` socket.

cosa65: Related reply [here](https://github.com/hms-dbmi-cellenics/ui/pull/885#discussion_r1327549119)

ivababukova (approved):

PR #347, opened 2023-10-14
use iac config for scaledown cron

alexvpickering:
# Description
Stage https://github.com/hms-dbmi-cellenics/worker/pull/347.

Also need to enable the cron task manually for biomage deployments.

# Details
#### URL to issue
N/A

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


chart-infra/templates/scale-to-zero.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if (eq .Values.kubernetes.env "staging") }}
+# workerScaleToZero isn't picked up - needs to be in master?
+{{- if or (eq .Values.myAccount.workerScaleToZero "true") (eq .Values.kubernetes.env "staging") }}
```
alexvpickering: `.Values.myAccount.workerScaleToZero` doesn't seem to be transmitted from `iac` correctly. Current guess is that `iac` changes need to be in master. May also be configured incorrectly

python/src/worker/config/__init__.py
```diff
@@ -7,15 +7,12 @@
...
-
-timeout = int(
-    os.getenv(
-        "WORK_TIMEOUT",
-        default=str(60 * 10),
-    )
-)
+default_timeout = 60 * 10
+timeout = os.getenv("WORKER_TIMEOUT", default = str(default_timeout))
+timeout = int(timeout) if timeout else default_timeout
```
alexvpickering: necessary for if `WORKER_TIMOUT` is empty string. `os.getenv(..., default=)` only uses default if variable is `None`

python/Dockerfile
```diff
@@ -1,13 +1,12 @@
 # pull official base image
-FROM python:3.8-buster AS builder
+FROM python:3.8-slim-buster AS builder
```
alexvpickering, line 2: smaller base image

python/Dockerfile
```diff
@@ -1,13 +1,12 @@
...
 
 # set working directory
 WORKDIR /src
 
 # install app dependencies
 COPY requirements.txt ./
-RUN pip3 install -r requirements.txt
-
+RUN pip3 install --no-cache-dir -r requirements.txt
```
alexvpickering, line 9: smaller layer

overall, compressed image for python container in ECR goes from ~630mb --> 100mb

python/requirements.txt
```diff
@@ -16,7 +16,6 @@ PyYAML==5.4
 requests==2.24.0
 responses==0.12.0
 sinfo==0.3.1
-umap-learn==0.4.6
```
alexvpickering, line 19: not needed on python image

chart-infra/templates/scale-to-zero.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if (eq .Values.kubernetes.env "staging") }}
+{{- if or (eq .Values.kubernetes.env "staging") (eq .Values.myAccount.accountId "160782110667") }}
```
alexvpickering, line 1: enabling cron for production in HMS

python/src/worker/config/domain_specific.py
```diff
@@ -0,0 +1,26 @@
...
+        return domain_specific['TEST']
+
+    aws_account_id = os.getenv('AWS_ACCOUNT_ID')
+
+    if aws_account_id == ACCOUNT_ID['HMS']:
+        return domain_specific['HMS']
+    elif aws_account_id == ACCOUNT_ID['BIOMAGE']:
+        return domain_specific['BIOMAGE']
+    else:
+        return domain_specific['BIOMAGE_PRIVATE']
```
alexvpickering: modeled off `api` `getDomainSpecificContent.js`

r/Dockerfile
```diff
@@ -73,14 +64,16 @@ ENV RENV_LIB=/src/lib
...
     cat sysdeps_run.txt | xargs ./install_debian_packages.sh && \
     rm -rf *
 
-# link system libstdc++ to conda installed version
-RUN rm /usr/lib/x86_64-linux-gnu/libstdc++.so.6 && \
-    ln -s /src/r-miniconda/envs/r-reticulate/lib/libstdc++.so.6 /usr/lib/x86_64-linux-gnu/libstdc++.so.6
+# install umap-learn for RunUMAP
+ENV WORKON_HOME=/src/.virtualenvs
+RUN R -q -e "reticulate::virtualenv_create('r-reticulate')" && \
+    R -q -e "reticulate::virtualenv_install('r-reticulate', 'umap-learn==0.5.3', pip_options='--no-cache-dir')"
```
alexvpickering, line 76: motivation is reducing image size (before: ~1010 MB, after: 884MB) which speeds up worker pod being ready. also a bit simpler

Together with python image size reduction (~530MB) worker ready time is down from ~2:30 --> 1:50

python/src/worker/config/domain_specific.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_domain_specific():
+    aws_account_id = os.getenv('AWS_ACCOUNT_ID')
+
+    if aws_account_id == ACCOUNT_ID['HMS']:
+        return domain_specific['HMS']
+    elif aws_account_id == ACCOUNT_ID['BIOMAGE']:
+        return domain_specific['BIOMAGE']
+    else:
+        return domain_specific['BIOMAGE_PRIVATE']
```
alexvpickering, line 23: modeled after `getDomainSpecificContent.js` in api

python/requirements.txt
gerbeldo: I just saw this dependency in an unrelated context and was confused by it. I suppose it's a vestigial remnant.

r/Dockerfile
gerbeldo: finally removing conda! I tried something similar months ago but got stuck. We can port this to the pipeline as well

python/requirements.txt
alexvpickering: yeah there are probably other that may not be needed either but doesn't seem so important since image size will be ~100MB now

r/Dockerfile
```diff
@@ -73,14 +64,16 @@ ENV RENV_LIB=/src/lib
...
+# python3-dev prevents 'reticulate can only bind to copies of Python built with --enable-shared'
 # cleanup setup files
-RUN echo "python3-pip" >> sysdeps_run.txt && \
+RUN echo "python3-pip python3-venv python3-dev" >> sysdeps_run.txt && \
     cat sysdeps_run.txt | xargs ./install_debian_packages.sh && \
     rm -rf *
 
-# link system libstdc++ to conda installed version
-RUN rm /usr/lib/x86_64-linux-gnu/libstdc++.so.6 && \
-    ln -s /src/r-miniconda/envs/r-reticulate/lib/libstdc++.so.6 /usr/lib/x86_64-linux-gnu/libstdc++.so.6
```
alexvpickering, line 83: don't think any equivalent of this is necessary anymore as libstdc++.so.6 isn't in the environment folder.

can't remember exactly what error under what set of conditions this fixed but `RunUMAP` works fine.

r/Dockerfile
```diff
@@ -73,14 +64,16 @@ ENV RENV_LIB=/src/lib
 RUN echo ".libPaths(c('$RENV_LIB', .libPaths()))" >> $(R RHOME)/etc/Rprofile.site
 
 # install runtime system deps
+# python3-dev prevents 'reticulate can only bind to copies of Python built with --enable-shared'
 # cleanup setup files
-RUN echo "python3-pip" >> sysdeps_run.txt && \
+RUN echo "python3-pip python3-venv python3-dev" >> sysdeps_run.txt && \
```
gerbeldo, line 69: why not declare these in the `sysdeps_build_debian.txt` instead of the dynamically generated `sysdeps_run.txt`?

gerbeldo: `python3-pip` is actually already installed

https://github.com/hms-dbmi-cellenics/worker/blob/f07d5a1bcdebeb70c860beef324c1a11952ffcb8/r/setup/sysdeps_build_debian.txt#L4C1-L5C1

alexvpickering: `sysdeps_build_debian.txt` are dependencies for **building** the R-packages. Once they have been built, the **run-time** system dependencies for R packages are determined  by `get_sysdeps_run.R`. This strategy reduces the overall image size. There are however some additional dependencies that are necessary for non R-package things:

- python3-pip for the development build (not needed for prod but installed here for code simplicity)
- python3-venv to setup the venv for `umap-learn`
- python3-dev to have a copy of python that `reticulate` can use

gerbeldo: I tried it in the pipeline PR and it does not work (a pip install fails due to not finding pip), for some reason I don't understand.

alexvpickering: I'll take a look in the pipeline and see if I can get it set up as well

alexvpickering: Not tested but builds for me: https://github.com/hms-dbmi-cellenics/pipeline/pull/340

gerbeldo: Thanks, I'll test the pipeline PR.

Regarding build time dependencies, are they removed downstream, after building the R packages? That relates to my initial question, why do we need python3-pip installed twice?

gerbeldo (approved):

r/Dockerfile
alexvpickering: Yes the `common` build doesn't start from the end of the `builder`, but rather from a fresh `rocker/r-ver`. It transfers over already built packages etc (`COPY --from=builder /src /src`) and then installs only the system dependencies that are needed to **run** packages (less hefty than packages needed to **build** all the R/python packages). The approach is a common one to reduce the size of Docker images.

chart-infra/templates/scale-to-zero.yaml
ivababukova: Can you explain what does this CRON do?

ivababukova: It is because I wonder if we also want it in the community instance

PR #350, opened 2023-11-21
Add correct reduction names in downloaded Seurat object

saracastel:
# Description
When downloading the rds file that contains the Seurat object, the dimensionality reduction name is invariably set to "umap" even in instances where t-SNE has been applied.
When the embeddings are calculated, converted to json, and sent to S3, the information on which method was used to generate them (which is in the column names) is lost. This PR is to use the reduction method from the request.

Staging: https://ui-sara-ui57-api22-worker21.scp-staging.biomage.net/

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/89

N/A

#### Links to any PRs or resources related to this PR
https://github.com/hms-dbmi-cellenics/ui/pull/953
https://github.com/hms-dbmi-cellenics/api/pull/510

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

- [x] Started end-to-end tests on the latest commit.

### Documentation updates

### Optional

### DIFF
--- a/r/R/embedding.R
+++ b/r/R/embedding.R
@@ -55,12 +55,6 @@ runEmbedding <- function(req, data) {
     tidyr::complete(cells_id = seq(0, max(data@meta.data$cells_id))) %>%
     dplyr::select(-cells_id)
 
-  # Add reduction method identifier
-  col_names <- colnames(df_embedding)
-  red_id <- ifelse(grepl("umap", col_names, ignore.case = T), 0, ifelse(grepl("tsne", col_names, ignore.case = T), 1, 2))
-  red_id <- setNames(red_id, col_names)
-  df_embedding <- dplyr::bind_rows(red_id, df_embedding)
-
   map2_fun <- function(x, y) {
     if (is.na(x)) {
       return(NULL)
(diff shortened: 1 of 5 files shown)


r/R/embedding.R
```diff
@@ -113,9 +113,9 @@ assignEmbedding <- function(embedding_data, data, reduction_method = "umap") {
   embedding <- embedding[cells_id + 1, ]
   rownames(embedding) <- colnames(data)
 
-  embedding_key = "UMAP_"
-  if(reduction_method == "tsne") {
-    embedding_key = "tSNE_"
+  embedding_key <- "UMAP_"
+  if (reduction_method == "tsne") {
+    embedding_key <- "tSNE_"
```
kafkasl: I don't like this if a lot from a style point of view. You could use a dictionary  and index it with the reduction method. e.g.
```
const reductionKeys = {
    'umap': 'UMAP_",
    'tsne': 'tSNE_'
}
```
and then do `embedding_key = reduction_keys[reduction_method]`

or if the lowecase `t` in `tSNE` is not needed just turn the method into uppercase and add the underscore something like: 

`embedding_key <- uppercase(reduction_method) + '_'`

kafkasl (approved):

r/R/embedding.R
gerbeldo: Agree on the named list! 

The low case "t" is the usual way t-SNE is stylized. I don't have strong opinions, but would check what Seurat does by default (by looking at the reduction slot of a Seurat object with a tsne embeddign).

Regarding the underscore, you can simplify the way you build the colnames below, with a single paste call with a 'sep="_"' argument, like:

colnames(embedding) <- paste(embedding_key, 1:2, sep = "_")

Paste coerces numbers to strings.

saracastel: Thanks both! I changed it a bit following your suggestions, but keeping tSNE with low case "t" as this is what Seurat does by default.

PR #357, opened 2024-03-15
update to R 4.4.0

alexvpickering:
# Description
- updates R to 4.4.0 and all packages
- updates code to reflect changes to Seurat (v5)

Things tested:
- [x] Default pipeline with 10x dataset
- [x] "within" and "between" DE with 10x and Seurat datasets
- [x] All integration methods with 10x dataset

staging:

https://ui-alex-worker357-pipeline363.staging.single-cell-platform.net

# Details
#### URL to issue
N/A

N/A

#### Links to any PRs or resources related to this PR
https://github.com/hms-dbmi-cellenics/pipeline/pull/363

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


r/R/cell_annotation_sctype.R
```diff
@@ -74,13 +72,13 @@ ScTypeAnnotate <- function(req, data) {
 #' @examples
 get_formatted_data <- function(scdata, active_assay) {
 
-  scale_data <- scdata[[active_assay]]@scale.data
+  scale_data <- scdata[[active_assay]]$scale.data
```
gerbeldo, line 75: here's what I meant in the pipeline question. do old seurat objects work with these changes? I assume they do, but just in case 😬

r/R/utilities_expression.R
```diff
@@ -113,7 +119,7 @@ completeExpression <- function(expression, cell_ids) {
     on = .(cell_ids)
     ]
 
-  expression[, cell_ids := NULL]
```
gerbeldo, line 116: why this change? in-place deletion is much much faster. And all this expression code needs to be as fast as it can be (data exploration user experience currently depends heavily on performance here, in particular to plot the gene expression values in the umap.)

r/R/utilities_expression.R
```diff
@@ -100,8 +107,7 @@ completeExpression <- function(expression, cell_ids) {
     return(expression)
   }
 
-  expression[, cell_ids := cell_ids]
```
gerbeldo, line 103: did you encounter any problems with in-place assignment here? I'm asking because I ran into some issues with the expression code a while ago.

r/R/utilities_expression.R
```diff
@@ -203,7 +209,7 @@ getStats <- function(data) {
 sparsify <- function(expression) {
   data.table::setnafill(expression, fill = 0)
   sparse_matrix <-
-    Matrix::Matrix(Matrix::as.matrix(expression), sparse = T)
+    as(as.matrix(expression), 'dgCMatrix')
```
gerbeldo, line 212: yesss that conversion was horrible.

r/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN echo ".libPaths(c('$RENV_LIB', .libPaths()))" >> $(R RHOME)/etc/Rprofile.sit
 
 # install renv to install required R packages
 RUN R -q -e "install.packages('remotes', repos = c(CRAN = 'https://cloud.r-project.org'))" && \
-    R -q -e "remotes::install_github('rstudio/renv@0.16.0')" && \
+    R -q -e "remotes::install_github('rstudio/renv@v1.0.7')" && \
```
gerbeldo, line 22: same suggestion, changing to PPM cran repo to speedup build.

gerbeldo (approved):
looks good!

r/Dockerfile
alexvpickering: single package/dependencies installation probably not worthwhile

r/R/cell_annotation_sctype.R
alexvpickering: Yes they do. Seurat V5 and V3 objects can be accessed in this manner (`scdata[[assay]]$layer`) while only Seurat V3 objects can be accessed in the old manner (`scdata[[assay]]@slot`)

r/R/utilities_expression.R
alexvpickering: Yes, there was some side effects outside of the function (something in the parent function was being updated unintentionally). 

This kind of operation in incredibly fast in base R -- no way that it limits the speed of this function at all.

r/R/utilities_expression.R
alexvpickering: Again, due to unintended side effects in parent function and because removal of a column is incredible fast (I would guess on order of nanoseconds). This is premature optimization -- other things will be slower in this function and so optimizing this will have no effect.

r/R/utilities_expression.R
alexvpickering: Changed this because some of the tests were creating a `dTiMatrix` or something like that which led to errors. This just makes it explicit the type of sparse matrix we want.

gerbeldo: yep, newer matrix converts (mathematically) diagonal matrices to that diagonal matrix type. The other solution is to add `diagonal = F` to the Matrix::Matrix call, but this is much nicer.

alexvpickering: as okay cool -- thanks for the heads up. We do have a few other calls of `Matrix::Matrix(..., sparse = T)`. They don't seem to be causing any errors in our tests but maybe they should be updated as well to prevent `dTiMatrix` shenanigans

